PR #2, opened 2020-09-22
Create the interface Agent and the classes Location, Board and BoardTest

OrpazFishel:


omermadmon:
Using javadoc tags when commenting will allow us to easily generate javadocs for the project, using [Apache Maven Javadoc Plugin](https://maven.apache.org/plugins/maven-javadoc-plugin/).
This [How to Write Doc Comments for the Javadoc Tool](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html) technical article seems to be useful.

src/main/java/com/google/research/bleth/simulator/Board.java
```diff
@@ -0,0 +1,58 @@
+package com.google.research.bleth.simulator;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+/** A container for all the agents, representing their locations, either real or estimated. */
+public class Board {
+    public final int rowNum;
+    public final int colNum;
+    ArrayList<Agent>[][] matrix;
```
nolaviz, line 10: Prefer to avoid Java's primitive array types - I would recommend Guava Table instead.
https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Table.html

Also see go/ej3e-28, http://doc/1LOhv9tywZmQ5shwhFDMpplpBoUiH142hoqSObCQPXKk/view

src/main/java/com/google/research/bleth/simulator/Board.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public Board(int rows, int cols) {
```
nolaviz, line 12: Please add JavaDoc (in particular - describe the state of a newly-created Board).

src/main/java/com/google/research/bleth/simulator/Board.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    /**
+     * Place an agent on board if the given location is valid.
+     * @param newLocation is the location where the agent will be placed.
+     * @param agent is the Agent which placed on the board.
+     */
+    public void placeAgent(Location newLocation, Agent agent) {
+        boolean isRowValid = newLocation != null && 0 <= newLocation.row && newLocation.row < rowNum;
```
nolaviz, line 27: Consider using Guava Preconditions to simplify the error checking here.
https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html

src/main/java/com/google/research/bleth/simulator/Board.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Remove an agent from its current location on board and place it on new location.
+     * @param oldLocation is the agent's current location.
+     * @param newLocation is the location where the agent will be placed.
+     * @param agent is the Agent which moves from oldLocation to newLocation.
+     */
+    public void moveAgent(Location oldLocation, Location newLocation, Agent agent) {
+        boolean isOldRowValid = oldLocation != null && 0 <= oldLocation.row && oldLocation.row < rowNum;
```
nolaviz, line 45: Prefer to factor out the validation of a Location into a helper method, instead of repeating it three times (once in placeAgent() and twice here).

src/main/java/com/google/research/bleth/simulator/Agent.java
```diff
@@ -0,0 +1,11 @@
+package com.google.research.bleth.simulator;
+
+/** Represent both observer and beacon. */
+public interface Agent {
+
+    /**
+     * Calculate the location the agent is moving to, based on its current location and its strategy.
+     * @return the location on board which the agent is moving to.
+     */
+    Location move();
```
nolaviz, line 10: It appears each Agent is aware of its own location; how do you ensure the Agent and any associated Board are truly in sync about this? (I expect we want to keep an invariant of "for every Board that has a reference to Agent _a_, _a_'s location on the board is the same as its internal notion of location"? Or is this sometimes broken, e.g. when the Board is an estimated one?)

src/main/java/com/google/research/bleth/simulator/Agent.java
```diff
@@ -0,0 +1,11 @@
+package com.google.research.bleth.simulator;
+
+/** Represent both observer and beacon. */
+public interface Agent {
```
nolaviz, line 4: Consider adding a getLocation() method so the agent can report its existing location.

src/main/java/com/google/research/bleth/simulator/Agent.java
nolaviz, line 10: Does this also change the agent's internal notion of location? Or is this done separately (how?)

src/main/java/com/google/research/bleth/simulator/Location.java
```diff
@@ -0,0 +1,12 @@
+package com.google.research.bleth.simulator;
+
+/** A location on board. */
+public class Location {
```
nolaviz, line 4: Consider using AutoValue (https://github.com/google/auto/blob/master/value/userguide/index.md).

Its main advantage for you - in this specific case - might be just the auto-generated toString() (otherwise debugging is a pain). But other cases may also need proper equals() and hashCode() semantics.

src/main/java/com/google/research/bleth/simulator/Board.java
```diff
@@ -0,0 +1,58 @@
+package com.google.research.bleth.simulator;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+/** A container for all the agents, representing their locations, either real or estimated. */
+public class Board {
+    public final int rowNum;
```
nolaviz, line 8: Avoid exposing internal fields publicly - prefer to provide accessors for them.
(Accessors enable you to change implementation details without affecting your callers - e.g., in this case, if you decide to accept my suggestion to use a Table, even the test should not know)

src/test/java/com/google/research/bleth/simulator/BoardTest.java
```diff
@@ -0,0 +1,307 @@
...
+        when(firstAgent.move()).thenReturn(zeroOnOneCoordinate);
+        when(secondAgent.move()).thenReturn(zeroOnOneCoordinate);
+
+        board.placeAgent(firstAgent.move(), firstAgent);
+        board.placeAgent(secondAgent.move(), secondAgent);
+
+        assertThat(board.matrix[0][1]).containsExactly(firstAgent, secondAgent);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
```
nolaviz, line 88: Prefer to use JUnit's assertThrows() instead of "expected" exceptions. See https://howtodoinjava.com/junit5/expected-exception-example/

(Note that use of "expected = FooException" in @Test is _strongly discouraged_ in Google).

Advantages of assertThrows():
* You can be sure the assertion was thrown in the specific statement, and not e.g. in some preparation step.
* You can assert further details about the thrown exception (e.g. verify the message).

Disadvantages of expected exceptions:
* If your test has any assertions after the statement that threw the exception, they won't be run at all (so the test will pass even if the assertions actually would have failed).

nolaviz (changes requested):

src/main/java/com/google/research/bleth/simulator/Agent.java
OrpazFishel: Both the real and estimated boards are instances of the Board class. However, this invariant should be kept w.r.t the real board only. It works as follows:

- Each agent is aware of its location on the real board.
- The agent's move method updates the agent's internal notion of location and returns this new location.
- The real board updates its matrix by calling moveAgent.

The resolver however doesn't use the agent's method move at all, it updates its board using the estimations based on the transmissions it receives.
The agent isn't aware at all to its location on the estimated board.

src/main/java/com/google/research/bleth/simulator/Location.java
omermadmon: We consider generating the Location class using protobuf (instead of this implementation), so these methods will be generated as well, and also this struct could be used in JS code (when parsing and visualizing a board state, for example).

nolaviz: Protobuf is also OK.

src/main/java/com/google/research/bleth/simulator/Agent.java
nolaviz: I still don't like this separation. It looks too fragile to me (as in: we have no guarantee that *only* the real Board will invoke Agent.move(), or we break an invariant - and we have no way to even detect that the invariant was broken).

Let's discuss alternatives?

OrpazFishel: Option 1:
Board will be an abstract class, we'll also have RealBoard and EstimatedBoard subclasses. The placeAgent() will be the only abstract method of Board. So now we have:

- Each agent is aware of its location on the real board.
- In addition, the Agent have a setRealLocation public method.
- The agent's move method returns this new location but does not update the agent's internal notion of location.
- The real board updates its matrix by calling placeAgent(someAgent.move(), someAgent).
- Only for RealBoard: Inside the placeAgent method the setRealLocation is being called to update the agent's internal notion of location.

This way both internal and global location on the real board are being updated in the same method.

Option 2:
The simulator itself will be responsible for both updating the real board and the agent's internal notion of location (using the board's placeAgent method and the agent's setter for location).

src/main/java/com/google/research/bleth/simulator/Agent.java
```diff
@@ -0,0 +1,21 @@
...
+     * Calculate the location the agent is moving to, based on its current location and its strategy.
+     * @return the location on board which the agent is moving to.
+     */
+    Location moveTo();
+
+    /**
+     * Move the agent to its next location and update the board accordingly.
+     */
+    void move();
+}
```
OrpazFishel, line 21: And the implementation will be something like:
    public void move() {
        Location nextMove = moveTo();
        simulation.getBoard().moveAgent(realLocation, nextMove, this);
        realLocation = nextMove;
    }

src/main/java/com/google/research/bleth/simulator/Board.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Check if a location is within the board boundaries.
+     * @param location is the location to check if valid.
+     * @return true if the location is valid, false otherwise.
+     */
+    public boolean isLocationInvalid(Location location) {
+        Preconditions.checkNotNull(location);
+        boolean isRowValid = 0 <= location.row && location.row < rowNum;
```
nolaviz, line 43: You could just use Preconditions.checkArgument() for these.

src/test/java/com/google/research/bleth/simulator/BoardTest.java
```diff
@@ -0,0 +1,380 @@
...
+
+        assertThat(board.getAgentsOnLocation(zeroOnOneCoordinate)).containsExactly(firstAgent, secondAgent);
+    }
+
+    @Test
+    public void placeNullOnBoardThrowsException() {
+        Board board = new Board(2, 2);
+        Mockito.when(firstAgent.moveTo()).thenReturn(zeroOnZeroCoordinate);
+
+        Assertions.assertThrows(NullPointerException.class, () -> {
```
nolaviz, line 94: Prefer to statically import assertThrows().

src/main/java/com/google/research/bleth/simulator/Board.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    /**
+     * Check if a location is within the board boundaries.
+     * @param location is the location to check if valid.
+     * @return true if the location is valid, false otherwise.
+     */
+    public boolean isLocationInvalid(Location location) {
+        Preconditions.checkNotNull(location);
```
nolaviz, line 42: Consider statically importing Preconditions.checkNotNull() (and any other methods you use from this class: their names are sufficiently clear to not require the class-name context).

src/main/java/com/google/research/bleth/simulator/Agent.java
nolaviz: Looks good.

nolaviz (approved):

PR #3, opened 2020-09-27
Add the interfaces and classes refer to Beacon and Movement Strategy

OrpazFishel:


src/main/java/com/google/research/bleth/simulator/RandomMovementStrategy.java
```diff
@@ -0,0 +1,36 @@
...
+     * @param board is the board that the agent is placed on.
+     * @param currentLocation is the current location of the agent on the board.
+     * @return the location on the board that the agent is moving to.
+     */
+    @Override
+    public Location moveTo(Board board, Location currentLocation) {
+        List<List<Integer>> shuffledDirections = new ArrayList<>(directions);
+        Collections.shuffle(shuffledDirections);
+        for (List<Integer> direction : shuffledDirections) {
+            Location newLocation = new Location(currentLocation.row + direction.get(0),
```
omermadmon, line 28: I assume this is not the only situation where we would like to add one location to another in order to generate a new location.

I wanted to suggest we use operators overloading to overload the '+' operator so we can just use:
```java
Location newLocation = currentLocation + direction;
```
But it seems like Java doesn't support operator overloading. 
Alternatively we can write a static 'add' method, and then use:
```java
Location newLocation = Location::Add(currentLocation, direction);
```

omermadmon: Okay I just noticed that 'direction' is not a Location but just a list of two integers. So we can either define a Direction class and then write a method to move a Location in a certain Direction, or just leave it as is.

src/main/java/com/google/research/bleth/simulator/StaticMovementStrategy.java
```diff
@@ -0,0 +1,16 @@
+package com.google.research.bleth.simulator;
+
+/** A static movement strategy for an agent. */
```
nolaviz, line 3: I'm not sure "static" is the best term here. Maybe "stationary" is preferred?

src/main/java/com/google/research/bleth/simulator/Beacon.java
```diff
@@ -0,0 +1,76 @@
+package com.google.research.bleth.simulator;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class Beacon implements IBeacon {
+    static int beaconId = 0; // used for generating unique id for each beacon.
```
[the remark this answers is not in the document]
nolaviz: This is still here (and now unused)...?

src/main/java/com/google/research/bleth/simulator/Beacon.java
```diff
@@ -0,0 +1,59 @@
...
+     * Create new Beacon with consecutive serial number.
+     * @param id is a unique ID.
+     * @param initialLocation is the location on board where the agent is placed.
+     * @param movementStrategy determines how the agent moves.
+     * @param simulation is the world the agent lives in.
+     */
+    Beacon(int id, Location initialLocation, MovementStrategy movementStrategy, Simulation simulation) {
+        this.id = id;
+        realLocation = initialLocation;
+        simulation.getBoard().placeAgent(realLocation,this);
```
nolaviz, line 23: Consider moving this to the factory: as I mentioned in an earlier comment - any logic called from a constructor might be dangerous.
(E.g., what will happen if Board.placeAgent() wants to print out the agent's details in a debug message? The agent is still half-constructed at this point, so anything referring to e.g. movementStrategy may produce an NPE...)

src/main/java/com/google/research/bleth/simulator/Board.java
```diff
@@ -38,21 +38,35 @@ public Board(int rows, int cols) {
     /**
      * Check if a location is within the board boundaries.
      * @param location is the location to check if valid.
-     * @return true if the location is valid, false otherwise.
+     * @throws IllegalArgumentException if the location is without the board boundaries.
```
nolaviz, line 41: You might also throw a NullPointerException if location is null.

src/main/java/com/google/research/bleth/simulator/Direction.java
```diff
@@ -0,0 +1,24 @@
+package com.google.research.bleth.simulator;
+
+public enum Direction {
```
nolaviz, line 3: Please add JavaDoc (to the class + its public members)

src/main/java/com/google/research/bleth/simulator/BeaconFactory.java
```diff
@@ -0,0 +1,24 @@
...
+     * Create new beacon, according to the given parameters.
+     * @param initialLocation is the location on board where the agent is placed.
+     * @param movementStrategy determines how the agent moves.
+     * @param simulation is the world the agent lives in.
+     */
+    public Beacon createBeacon(Location initialLocation, MovementStrategy movementStrategy, Simulation simulation) {
+        checkNotNull(initialLocation);
+        checkNotNull(movementStrategy);
+        checkNotNull(simulation);
+        if (!simulation.getBoard().isLocationValid(initialLocation)) {
```
nolaviz, line 19: Why not just invoke simulation.getBoard().validateLocation(initialLocation)?
It will throw the same exception...

nolaviz (approved):

PR #4, opened 2020-09-29
Data Serialization, Storage and Visualization

omermadmon:
* Created BoardParser, a utility class for Board objects parsing.
* Created DatabaseService, a class for interacting with the GC Datastore instance.
* Added unit testing classes.
* Added some servlets and JavaScript code for writing hard coded data into the Datastore instance, and visualizing board states.

image


src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,158 @@
+package com.google.research.bleth.services;
+
+import com.google.appengine.api.datastore.*;
```
OrpazFishel, line 3: Try to avoid wildcard imports.
https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Create an entity of a real baord state based on input params and write it to datastore.
+     * @param simulationId is the Id of the simulation associated with the input board state.
+     * @param round is the round number in the simulation that matches the input board state.
+     * @param boardState is a JSON string represnting the state of the board.
+     *                   Is parsable to a 3D array of strings (representing the agents' Ids).
+     * @throws PreparedQuery.TooManyResultsException if a board state with the same simulationId and round already exists.
+     */
+    public void writeRealBoardState(String simulationId, int round, String boardState)
```
OrpazFishel, line 61: Consider mentioning it's for tracing simulation at the method name/the Javadoc.

src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,158 @@
...
+        writeBoardState(simulationId, round, boardState, "TracingEstimatedBoardState");
+    }
+
+    private String getBoardState(String simulationId, int round, String queryKind)
+            throws PreparedQuery.TooManyResultsException {
+
+        // Set simple predicates.
+        Query.FilterPredicate p1 =
+                new Query.FilterPredicate("simulationId", Query.FilterOperator.EQUAL, simulationId);
+        Query.FilterPredicate p2 =
```
OrpazFishel, line 85: I prefer explicit names such as filterByRound instead of p2.

src/main/java/com/google/research/bleth/servlets/ReadHardCodedEstimatedBoardStateServlet.java
```diff
@@ -0,0 +1,35 @@
+package com.google.research.bleth.servlets;
+
+import com.google.research.bleth.services.DatabaseService;
+import com.google.research.bleth.simulator.Board;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
```
OrpazFishel, line 10: The imports within each block should appear in ASCII sort order. https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing

src/main/java/com/google/research/bleth/simulator/Board.java
```diff
@@ -1,15 +1,17 @@
 package com.google.research.bleth.simulator;
 
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkNotNull;
-
 import com.google.common.collect.ArrayTable;
 import com.google.common.collect.ImmutableList;
+import com.google.gson.Gson;
+
```
OrpazFishel, line 6: Unnecessary blank line.
https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing

src/main/java/com/google/research/bleth/simulator/Board.java
```diff
@@ -36,7 +38,7 @@ public Board(int rows, int cols) {
     }
 
     /**
-     * Check if a location is within the board boundaries.
+     * Check if a location is within the board boundaries. If not - throw an exception.
```
OrpazFishel, line 41: FYI - I've changed this function & Javadoc :)

src/main/java/com/google/research/bleth/simulator/DummyBeacon.java
```diff
@@ -0,0 +1,38 @@
+package com.google.research.bleth.simulator;
+
+/**
+ * A CLass used for testing board visualization.
```
OrpazFishel, line 4: Nit: CLass.

src/main/webapp/board_state_visualization.js
```diff
@@ -0,0 +1,97 @@
...
+        params.append('simulationId', 'demo-tracing-sim-1');
+        params.append('simulationId', simulationId);
+        params.append('round', roundNumber);
+
+        // Fetch and display real board state.
+        fetch('/read-hard-coded-real-board-state', {method: 'POST', body: params})
+        .then(response => response.json())
+        .then(boardState => visualizeHardCodedBoardState(boardState.array, true));
+
+        // Fetch and display real board state.
```
OrpazFishel, line 28: estimated board

src/main/webapp/board_state_visualization.js
```diff
@@ -0,0 +1,97 @@
...
+ */
+function sleep(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+
+/**
+ * Visualize the board state on index.html page.
+ * @param {boardState} board a JSON string representing the board state in a certain round of the simulation.
+ * @param {boolean} isReal   if 'true' visualize the real board, otherwise visualize the estimated board.
```
OrpazFishel, line 50: Nit: too many spaces.

src/main/webapp/board_state_visualization.js
```diff
@@ -0,0 +1,97 @@
...
+ */
+function visualizeHardCodedBoardState(board, isReal) {
```
OrpazFishel, line 52: I'll quote Alon:
"Google best practices recommend to avoid boolean parameters as much as possible; instead, we recommend to use two separate methods, whose names reflect the effect of the boolean, and possibly sharing aspects of the implementation. (The reasons to avoid boolean parameters are (a) they're hard to extend, and (b) they more-or-less guarantee higher complexity - more "if"s - within the method)."

src/main/webapp/board_state_visualization.js
```diff
@@ -0,0 +1,97 @@
...
+    // Set table and table body
+    if (isReal) { var gameBoardTable = document.getElementById('real-game-board'); }
+    else { var gameBoardTable = document.getElementById('estimated-game-board'); }
+    var gameBoardTableBody = document.createElement('tbody');
+
+    // Clear last state of table
+    gameBoardTable.innerHTML = '';
+
+    board.forEach(function(rowData) {
+
```
OrpazFishel, line 63: Nit: unnecessary blank line.

src/main/webapp/board_state_visualization.js
```diff
@@ -0,0 +1,97 @@
...
+ * @param {agents list} agents a list of agents ids located in the same cell of the board.
+ */
+function determineColor(agents) {
+    if (agents.length === 0) { return 'white'; }
+    var hasBeacon = false;
+    var hasObsrver = false;
+    for (var i = 0; i < agents.length; i++) {
+        if (agents[i].charAt(0) === 'B') { hasBeacon = true; }
+        if (agents[i].charAt(0) === 'O') { hasObsrver = true; }
+        if (hasBeacon) { return 'red'; }
```
OrpazFishel, line 93: If the first "if" statement returns "red" we won't need the flags nor the other "if" statements - if agents isn't empty and we don't return "red" it means there's at least one observer in agents.

src/test/java/com/google/research/bleth/services/DatabaseServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+
+    @Before
+    public void setUp() {
+        helper.setUp();
+    }
+
+    @Test
+    public void writeAndThenReadEmptyRealBoardState_ShouldGetEqualBoardState() {
+
+        DatabaseService db = DatabaseService.getInstance();
```
OrpazFishel, line 53: Why isn't it a class attribute? Isn't it the same db anyway?

src/test/java/com/google/research/bleth/services/DatabaseServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+        assertThat(emptyRealBoardState).isEqualTo(outputRealBoardState);
+    }
+
+    @Test
+    public void writeAndThenReadNonRealBoardState_ShouldGetEqualBoardState() {
+
+        DatabaseService db = DatabaseService.getInstance();
+        Board nonEmptyRealBoard = new Board(2, 2);
+        Mockito.when(firstAgent.moveTo()).thenReturn(zeroOnZeroCoordinate);
+        nonEmptyRealBoard.placeAgent(firstAgent.moveTo(), firstAgent);
```
OrpazFishel, line 68: You can just place firstBeacon at the wanted location on the board with placeAgent, there's no need to use Mockito or moveTo for that.

src/test/java/com/google/research/bleth/services/DatabaseServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+    @After
+    public void tearDown() {
+        helper.tearDown();
+    }
+
+    @AfterAll
+    static void clearDatastore() {
+        DatabaseService db = DatabaseService.getInstance();
+        db.deleteAllSimulationRealBoardStates(firstSimulationId);
+        db.deleteAllSimulationEstimatedBoardStates(firstSimulationId);
```
OrpazFishel, line 262: Isn't there any boardState with secondSimulationId?

omermadmon: Correct, but I just figured out that this method is not needed at all - thanks to the helper and the tearDown() method (the helper provides a local datastore service which is designated for testing only):

https://cloud.google.com/appengine/docs/standard/java/tools/localunittesting

src/main/java/com/google/research/bleth/servlets/WriteHardCodedBoardStatesServlet.java
```diff
@@ -0,0 +1,73 @@
...
+        datastore.deleteAllSimulationRealBoardStates(simulationId);
+        datastore.deleteAllSimulationEstimatedBoardStates(simulationId);
+
+        for (int round = 0; round < 10; round++) {
+            // Clear boards.
+            realBoard = new Board(5, 5);
+            estimatedBoard = new Board(5, 5);
+
+            // Randomly locate dummy beacons on both boards.
+            realBoard.placeAgent(new Location((int) (Math.random() * 3),(int) (Math.random() * 3)), firstDummyBeacon);
```
nolaviz, line 58: Ouch. I don't like: you just placed the beacon in a random location on the real board, that's different from the beacon's internal location. So - we created an inconsistent state.

Please use the `move` methods so the beacon's internal state reflects the real board.

src/main/java/com/google/research/bleth/simulator/Board.java
```diff
@@ -1,15 +1,16 @@
...
-
 import com.google.common.collect.ArrayTable;
 import com.google.common.collect.ImmutableList;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import static com.google.common.base.Preconditions.checkArgument;
```
nolaviz, line 11: Static imports should be before dynamic imports.
https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing

src/main/java/com/google/research/bleth/simulator/DummyBeacon.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * A Class used for testing board visualization.
+ */
+public class DummyBeacon implements Agent {
+
+    private final int id;
+    private Location realLocation;
+
+    public DummyBeacon(int id, Location realLocation) {
```
nolaviz, line 11: Consider making the `Beacon` aware of its `RealBoard` owner (so it will update its own location on the real board whenever its internal location changes).

src/main/java/com/google/research/bleth/servlets/WriteHardCodedBoardStatesServlet.java
```diff
@@ -0,0 +1,73 @@
...
+    public void init() {
+        realBoard = new Board(5, 5);
+        estimatedBoard = new Board(5, 5);
+        firstDummyBeacon = new DummyBeacon(0, new Location((int) (Math.random() * 3),(int) (Math.random() * 3)));
+        secondDummyBeacon = new DummyBeacon(1, new Location((int) (Math.random() * 3),(int) (Math.random() * 3)));
+        thirdDummyBeacon = new DummyBeacon(2, new Location((int) (Math.random() * 3),(int) (Math.random() * 3)));
+    }
+
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
nolaviz, line 45: As this servlet _does_ modify state, recommended practice is to use a `POST`.

src/main/webapp/board_state_visualization.js
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Visualize the estimated board state on index.html page.
+ * @param {boardState} board a JSON string representing the board state in a certain round of the simulation.
+ */
+function visualizeHardCodedEstimatedBoardState(board) {
+    visualizeHardCodedBoardState(board, 'estimated-game-board');
+}
+
+/**
+ * return 'white' is cell is empty, 'blue' if contains observers only and 'red' otherwise.
```
nolaviz, line 95: Typo: "is cell" --> "if cell"

src/main/webapp/board_state_visualization.js
```diff
@@ -0,0 +1,104 @@
...
+ * @param {agents list} agents a list of agents ids located in the same cell of the board.
+ */
+function determineColor(agents) {
```
nolaviz, line 98: Consider using CSS styles rather than explicit colors.

src/test/java/com/google/research/bleth/services/BoardParserTest.java
```diff
@@ -0,0 +1,137 @@
...
+    private static final Location oneOnOneCoordinate = new Location(1, 1);
+    private static final Location zeroOnOneCoordinate = new Location(0, 1);
+    private static final Location oneOnZeroCoordinate = new Location(1, 0);
+
+    @Mock
+    private Agent firstAgent;
+    @Mock
+    private Agent secondAgent;
+
+    public ArrayTable<Integer, Integer, ArrayList<String>> emptyStringListTable() {
```
nolaviz, line 34: Prefer to move helpers below the actual testcases: the testcases are more important to a reader of this class.

Also - helpers should be private. (Always minimize visibility of all elements in your code...)

src/test/java/com/google/research/bleth/services/BoardParserTest.java
```diff
@@ -0,0 +1,137 @@
...
+        for (int row = 0; row < 2; row++) {
+            for (int col = 0; col < 2; col++) {
+                table.set(row, col, new ArrayList<String>());
+            }
+        }
+
+        return table;
+    }
+
+    public void addAgentIdInLocation (ArrayTable<Integer, Integer, ArrayList<String>> table, Agent agent, Location location) {
```
nolaviz, line 48: Remove extra space before the "(".
Also - the method name is a bit awkward: "add" may not be the right verb here - maybe "place"?

src/test/java/com/google/research/bleth/services/BoardParserTest.java
```diff
@@ -0,0 +1,137 @@
...
+        table.get(location.row, location.col).add(agent.getType() + agent.getId());
+    }
+
+    @Test
+    public void parseEmptyBoard() {
+
+        // Arrange board.
+        Board board = new Board(2, 2);
+
+        // Arrange expected serialized board.
```
nolaviz, line 58: This is not the best way to test: you're verifying too much (e.g. the specific whitespace and formatting of the JSON). In effect, your test becomes a change detector - https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html

Prefer instead to parse the returned JSON, and verify that the result is *semantically* empty (i.e. all of the elements are empty lists).

The same comment applies to other tests as well.

src/test/java/com/google/research/bleth/services/DatabaseServiceTest.java
nolaviz: It's better practice to reduce the number of class attributes in a test class - we *encourage* repetition in tests (each testcase should be self-contained as far as possible).

src/test/java/com/google/research/bleth/services/DatabaseServiceTest.java
```diff
@@ -0,0 +1,247 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@RunWith(MockitoJUnitRunner.class)
+public final class DatabaseServiceTest extends TestCase {
+
+    private final LocalServiceTestHelper helper =
+            new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig()
+                    .setAutoIdAllocationPolicy(LocalDatastoreService.AutoIdAllocationPolicy.SEQUENTIAL));
+
+    private static final String firstSimulationId = "test-tracing-sim-id-1";
```
nolaviz, line 29: Constants should be ALL_UPPERCASE.
https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

src/test/java/com/google/research/bleth/services/DatabaseServiceTest.java
```diff
@@ -0,0 +1,247 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import static com.google.common.truth.Truth.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@RunWith(MockitoJUnitRunner.class)
+public final class DatabaseServiceTest extends TestCase {
```
nolaviz, line 23: I got lost in the scenarios here... Are these complex scenarios really necessary?

Why do you need more than
* Write a specific (non-empty) real board, verify it reads back correctly
* Same for estimated
* Verify deletion works

Do we really care about these "multiple simulation ID" cases? Can they ever happen in practice?

nolaviz (changes requested):

src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Create an entity of a real board state based on input params and write it to datastore.
+     * @param simulationId is the Id of the simulation associated with the input board state.
+     * @param round is the round number in the simulation that matches the input board state.
+     * @param board is a Board object representing the state of the board.
+     * @throws PreparedQuery.TooManyResultsException if a board state with the same simulationId and round already exists.
+     */
+    public void writeRealBoardState(String simulationId, int round, Board board)
```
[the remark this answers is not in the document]
omermadmon: This sounds like a great solution for the write method, I'm on it.

But as for reading I believe we will still have two separate methods, because we will most likely have both real and estimated entities for the same combination of simulation ID and round number stored on the db.

An alternative could be having a single method that gets a flag for real or estimated as an argument, but I recall you've already mentioned this solution as a bad practice.

src/main/java/com/google/research/bleth/simulator/DummyBeacon.java
omermadmon: The dummy beacon always stays at the same location, therefore I didn't think it was necessary (the actual beacons and observers are aware of their `Simulation` owner, and therefore their owner `Real Board` as well).

src/main/java/com/google/research/bleth/servlets/WriteHardCodedBoardStatesServlet.java
omermadmon: Oh my.. exactly the kind of mistakes I'm having on exams :(

src/main/java/com/google/research/bleth/services/DatabaseService.java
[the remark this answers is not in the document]
OrpazFishel: Is an attribute of typeBoard (enum which can be REAL or ESTIMATED) good enough for this purpose? 
Is there any benefit in making Board abstract as you suggested?

src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Retrieve the real board state who matches the input params (if exists) and return it as a JSON string.
+     * @param simulationId is the Id of the simulation.
+     * @param round is the round number in the simulation.
+     * @return a JSON string representing the corresponding board state, or null if it doesn't exist.
+     * @throws PreparedQuery.TooManyResultsException if there are more than a single entity matching the input params.
+     */
+    public String getRealBoardState(String simulationId, int round)
```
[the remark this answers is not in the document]
omermadmon: The `write` method gets a board and write to the db a JSON string.
The `read` method gets a simulation ID and round number and retrieves the JSON string representing this board state.

The reason `read` returns a JSON string and not a Board object is that the string can be written to a `HTTPServletResponse` and it is the format the UI code expects to receive in order to visualize the board state:

```java
// A servlet for reading a board state.
String boardState = DatabaseService.getInstance().readRealBoardState(simulationId, round);
response.getWriter().println(boardState);
```

If we change the `read` method return value to be a Board, we will have to add another phase of translating from Board to JSON string outside the `read` method:

```java
// A servlet for reading a board state.
Board boardStateAsBoard= DatabaseService.getInstance().readRealBoardState(simulationId, round);
String boardStateAsBoardJson = BoardParser.encode(boardStateAsBoard) // <---
response.getWriter().println(boardStateAsBoardJson);
```
I personally prefer the first implementation (unbalanced, but requires less operations inside the servlet), please let me know if there are any other advantages of the second one.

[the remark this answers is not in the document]
nolaviz: There are two issues with this approach:
1. It hardcodes the assumption that the database format is one-and-the-same as the communication format. If we ever decide to vary them, even slightly - this entire structure breaks. (Google best practices call for e.g. using different protos at these layers - precisely because changes do happen, sometimes, and if you're not ready for them the change becomes extra-costly.)
2. If we ever implement additional logic in the server that reads from the database (e.g. "continue a saved simulation for another 1000 steps") - the JSON-returning read() makes this harder than it should be.

src/test/java/com/google/research/bleth/services/DatabaseServiceTest.java
nolaviz: I agree we may need more than just the 3 simple scenarios I listed, but - you still have far too many IMO.
For example, there's no real need to verify every combination of parameters - a single testcase can verify that changing simulation ID is enough to get a different board; another testcase verifies that changing just the round number also gets a different board; and that's it.

src/main/java/com/google/research/bleth/services/DatabaseService.java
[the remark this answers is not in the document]
omermadmon: Do we really want to return a `Board` without an associated `Simulation` object? 

If we do: this means that when creating the returned `Board` we are creating `Agent` objects with default movement and awakeness strategies (e.g stationary), and a `null` value where should be a reference to the `Simulation` object.

If we don't: that means we need to store in the db all information required for re-generating the `Simulation` object, including:
- Agent's movement stratregies.
- Observers' awakeness strategies.
- Resolver current mapping state (which is not only the estimated board state, but also the transmission received in in the last round).

In such case, the `DatabaseService` class will have to be updated whenever a new simulation class is being pushed to the master, and also a [JsonSeriallizer](https://www.javadoc.io/doc/com.google.code.gson/gson/2.8.5/com/google/gson/JsonSerializer.html) and a [JsonDeserializer](https://www.javadoc.io/doc/com.google.code.gson/gson/2.8.5/com/google/gson/JsonDeserializer.html) should be created for each class we want to be able to write to the db.

Most of this additional data to be stored is not relevant for visualization, but only for creating the returned board. Moreover, if we return a board associated to a simulation, when and where this simulation is created?

PR #5, opened 2020-10-01
Add the interfaces and classes refer to Observer and Awakeness Strategy 

OrpazFishel:


src/main/java/com/google/research/bleth/simulator/IObserver.java
```diff
@@ -0,0 +1,19 @@
...
+     */
+    void observe(Transmission beaconTransmission);
+
+    /**
+     * Deliver all the transmissions the observer observed in the current round to its resolver.
+     */
+    void passInformationToResolver();
+
+    /** Returns true if the observer is active, false otherwise. */
+    boolean isObserverAwake();
```
omermadmon, line 18: I my opinion
```java
observer.isAwake();
```
Would be better than
```java
observer.isObserverAwake();
```

Because it is already clear that this methods refers to an observer (but up to you).

src/main/java/com/google/research/bleth/simulator/IObserver.java
```diff
@@ -0,0 +1,19 @@
+package com.google.research.bleth.simulator;
+
+/** Represent an observer. */
+public interface IObserver extends Agent {
```
nolaviz, line 4: Unclear naming: some of your interfaces just use an unadorned name ("Agent"), some - like this one - add an "I" prefix.
Google naming guidelines recommend the unadorned names.
In any case, please be consistent.

src/main/java/com/google/research/bleth/simulator/Observer.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Activate the observer if the current round is the start of its current awakeness time
+     * and turn it off if it's the end of its current awakeness time.
+     * The function is called for every round between 1 and the last round without skipping.
+     * @param currentRound is the the current round of the simulation.
+     */
+    public void updateAwakenessState(int currentRound) {
+        if (awake && currentRound == nextAwakeningTime + awakenessDuration) {
```
nolaviz, line 99: Consider changing this to `>=`, so it will work even if the simulation does skip a round.

src/main/java/com/google/research/bleth/simulator/RandomAwakenessStrategy.java
```diff
@@ -0,0 +1,24 @@
...
+     * @param nextIntervalStart is the first round in which the observer can wake up again.
+     * @param awakenessCycle is the duration of each awakeness cycle - an observer can wake up once such cycle.
+     * @param awakenessDuration is the number of rounds that the observer is awake each time it wakes up.
+     * @param lastAwakeningTime is the last round the observer woke up.
+     * @return the next round in which the observer wakes up.
+     */
+    @Override
+    public int nextTime(int nextIntervalStart, int awakenessCycle, int awakenessDuration, int lastAwakeningTime) {
+        if (awakenessCycle == awakenessDuration) {
+            return 0;
```
nolaviz, line 20: Is this correct? You can return a value below nextIntervalStart - seems to me like this will break the logic in updateAwakenessState?
(Consider if this may mean something for the API: e.g., maybe the awakeness strategy should return a positive sleep duration, rather than an absolute wakeup cycle?)

src/test/java/com/google/research/bleth/simulator/ObserverTest.java
```diff
@@ -0,0 +1,930 @@
...
+    private static final BeaconFactory beaconFactory = new BeaconFactory();
+    private static final ObserverFactory observerFactory = new ObserverFactory();
+
+    @Mock
+    private Simulation simulation;
+
+    /**
+     * A fake implementation for IResolver that allows to test the arguments it receives when an observer calls receiveInformation.
+     */
+    private class FakeResolver implements IResolver {
```
nolaviz, line 29: Consider splitting this to a separate file (even if it's not going to be reused - just so it doesn't clutter the tests file).

src/test/java/com/google/research/bleth/simulator/ObserverTest.java
```diff
@@ -0,0 +1,930 @@
...
+
+        @Override
+        public Board getBoard() {
+            return null;
+        }
+    }
+
+    private IResolver resolver = new FakeResolver();
+
+    private Observer createStaticObserverOnLocation(Location initial_location) {
```
nolaviz, line 55: Prefer to place helper methods below the tests: they make more sense once someone already knows what they're used for. (Also, the most important part of a test class are the tests - you want a reader to get to them early.)

src/test/java/com/google/research/bleth/simulator/ObserverTest.java
```diff
@@ -0,0 +1,930 @@
...
+    private boolean isMovementOfCorneredObserverIsValid(Board board, Location corner) {
+        Mockito.when(simulation.getBoard()).thenReturn(board);
+        Observer randomObserver = createRandomObserverOnLocation(corner);
+
+        Location nextLocation = randomObserver.moveTo();
+
+        return simulation.getBoard().isLocationValid(nextLocation) && calculateDistance(corner, nextLocation) <= 1;
+    }
+
+    private boolean movingObserverLocationIsUpdatedOnBoardMatrix(Board board, Location initialLocation) {
```
nolaviz, line 87: This is a helper, but it has enough logic (creating objects/modifying/asserting) that it seems it should just be a test on its own.

src/test/java/com/google/research/bleth/simulator/ObserverTest.java
```diff
@@ -0,0 +1,930 @@
...
+
+    private Beacon createStaticBeaconOnLocation(Location initial_location) {
+        return beaconFactory.createBeacon(initial_location, new StationaryMovementStrategy(), simulation);
+    }
+
+    private int calculateDistance(Location oldLocation, Location newLocation) {
+        return Math.abs(newLocation.row - oldLocation.row) + Math.abs(newLocation.col - oldLocation.col);
+    }
+
+    private boolean isMovementOfCorneredObserverIsValid(Board board, Location corner) {
```
nolaviz, line 78: This is a helper, but it has enough logic (creating objects/modifying/asserting) that it seems it should just be a test on its own.

src/test/java/com/google/research/bleth/simulator/ObserverTest.java
```diff
@@ -0,0 +1,930 @@
...
+        Mockito.when(simulation.getBoard()).thenReturn(board);
+        Observer staticObserver = createStaticObserverOnLocation(zeroOnZeroCoordinate);
+
+        staticObserver.move();
+
+        assertThat(staticObserver.getLocation()).isEqualTo(zeroOnZeroCoordinate);
+    }
+
+    @Test
+    public void randomObserverSurroundedByFourDirectionsStayOnItsLocation() {
```
nolaviz, line 222: This seems like a duplicate of `randomObserverSurroundedByFourDirectionsNextMoveIsToItsLocation`? (Same applies to the next few tests)

src/test/java/com/google/research/bleth/simulator/ObserverTest.java
```diff
@@ -0,0 +1,930 @@
...
+        Mockito.when(simulation.getBoard()).thenReturn(board);
+        Observer randomObserver = createRandomObserverOnLocation(zeroOnOneCoordinate);
+
+        randomObserver.move();
+
+        assertThat(randomObserver.getLocation()).isEqualTo(zeroOnZeroCoordinate);
+    }
+
+    @Test
+    public void randomObserverMoveExactlyOneStep() {
```
nolaviz, line 296: If it's a random observer, in theory this test may be "flaky": some moves will be OK, some not.
Consider instead testing e.g. 1000 consecutive moves, and verifying all of them are OK. (This just reduces uncertainty, but with random - that's all we can hope for.)

src/test/java/com/google/research/bleth/simulator/ObserverTest.java
```diff
@@ -0,0 +1,930 @@
...
+        Mockito.when(simulation.getBoard()).thenReturn(board);
+        Observer randomObserver = createRandomObserverOnLocation(oneOnOneCoordinate);
+
+        randomObserver.move();
+
+        assertThat(calculateDistance(randomObserver.getLocation(), oneOnOneCoordinate)).isEqualTo(1);
+    }
+
+    @Test
+    public void randomObserverSurroundedByThreeDirectionsMoveExactlyOneStepAside() {
```
nolaviz, line 307: Do you really need both this test and the one with a `Board(1, 2)`? This test covers both behaviors - (a) observer will move to the right, and (b) it will move just a single step.

src/test/java/com/google/research/bleth/simulator/ObserverTest.java
```diff
@@ -0,0 +1,930 @@
...
+        randomObserver.observe(firstBeaconTransmission);
+        randomObserver.passInformationToResolver();
+        randomObserver.observe(secondBeaconTransmission);
+        randomObserver.passInformationToResolver();
+
+        assertThat(((FakeResolver) resolver).getTransmissions()).containsExactly(secondBeaconTransmission);
+    }
+
+    @Test
+    public void fixedAwakenessObserverWithFirstAwakenessTimeZeroStartsAwake() {
```
nolaviz, line 644: If you perform the change I had suggested on AwakenessStrategy (move its parameters inside), you could split off this set of tests and just verify the AwakenessStrategy directly.

Furthermore - you could implement the tests as an abstract test class that then applies to multiple concrete awakeness strategies (so you write the tests once, then run the tests on all implementations).

nolaviz (changes requested):

src/test/java/com/google/research/bleth/simulator/ObserverTest.java
OrpazFishel: One is for moveTo and the second is for move, but I guess that testing move is enough because it uses moveTo.

src/main/java/com/google/research/bleth/simulator/AbstractAgent.java
```diff
@@ -0,0 +1,40 @@
...
+     */
+    @Override
+    public Location moveTo() {
+        return IMovementStrategy.moveTo(owner, realLocation);
+    }
+
+    @Override
+    public void move() {
+        Location nextMove = moveTo();
+        owner.updateAgentLocation(realLocation, nextMove,this);
```
nolaviz, line 37: Nit: missing space before "this"

src/test/java/com/google/research/bleth/simulator/AwakenessStrategyTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+import org.junit.runner.RunWith;
+import org.junit.Test;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public abstract class AwakenessStrategyTest {
+    @Test
+    public void observerWithFirstAwakenessTimeZeroStartsAwake() {
+        IAwakenessStrategy awakenessStrategy = createAwakenessStrategy(5, 3, 0);
```
nolaviz, line 13: Google best practices recommend to document ambiguous literal arguments at the call site (i.e., any argument whose meaning isn't obvious - due to being a named constant, variable, etc.)

So this would become

```
  createAwakenessStrategy(
      /* awakenessCycleDuration= */ 5, 
      /* awakenessDuration= */ 3,
      /* firstAwakenessTime= */ 0);
```

(applies to all other tests as well, obviously)

[Google internal tooling would also report an error if the parameter name does not match the comment, but we don't have that in Github :-( ]

src/test/java/com/google/research/bleth/simulator/AwakenessStrategyTest.java
```diff
@@ -0,0 +1,102 @@
+package com.google.research.bleth.simulator;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import org.junit.runner.RunWith;
+import org.junit.Test;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public abstract class AwakenessStrategyTest {
```
nolaviz, line 10: Having an abstract test for an abstract class is exactly right.

src/test/java/com/google/research/bleth/simulator/IAgentTest.java
```diff
@@ -0,0 +1,262 @@
+package com.google.research.bleth.simulator;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import org.junit.runner.RunWith;
+import org.junit.Test;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public abstract class IAgentTest {
```
nolaviz, line 10: Same as with AwakenessStrategyTest - having an abstract test class for an interface is again the right choice.

nolaviz (approved):

PR #6, opened 2020-10-07
Simulation and SimulationBuilder classes and subclasses

omermadmon:
- Added `Simulation` abstract class, with `SimulationBuilder` and `SimulationBuilderFromExisting` inner static abstract classes.
- Added concrete classes: `TracingSimulation`, `TracingSimulationBuilder` and `TracingSimulationBuilderFromExisting`.
- Added test cases to demonstrate the creation of a `TracingSimulation` using each builder class.

The `SimulationBuilder` will be used whenever a new simulation should be created from scratch, while the `SimulationBuilderFromExisting` will be used when we want to initialize the simulation state based on existing data (e.g after reading from db).

I have also manually added some interface and classes implementation without any logic to keep the code readable and compilable.


src/main/java/com/google/research/bleth/simulator/Simulation.java
```diff
@@ -1,9 +1,380 @@
...
+ */
 public abstract class Simulation {
-    Board board;
 
-    public Board getBoard() {
+    private final String id;
+    protected int currentRound = 0;
+    private final int maxNumberOfRounds;
+    private Board board;
+    protected ArrayList<Beacon> beacons;
```
OrpazFishel, line 16: If we don't change beacons during the Simulation, prefer to use ImmutableList to store them.

src/main/java/com/google/research/bleth/simulator/Simulation.java
```diff
@@ -1,9 +1,380 @@
...
+     */
+    void observersToResolver() { }
+
+    /**
+     * Update resolver's estimated board.
+     */
+    void resolverEstimate() { }
+
+    /**
+     * Write current-round real and estimated board states to db.
```
OrpazFishel, line 87: current-round state of the whole Simulation

src/main/java/com/google/research/bleth/simulator/Simulation.java
```diff
@@ -1,9 +1,380 @@
...
+         */
+        public SimulationBuilder setRadius(double radius) {
+            this.radius = radius;
+            return this;
+        }
+
+        /**
+         * Validate all simulation builder arguments are legal.
+         */
+        public abstract void validateArguments();
```
OrpazFishel, line 225: Why does the asccess modifier is public?

src/main/java/com/google/research/bleth/simulator/TracingSimulation.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * A simulation in which the global resolver estimates beacons real locations based on information received from observers.
+ * Beacons transmit their unique static IDs each round.
+ * Agents move according to a movement strategy.
+ * Observers change their awakeness states according to an awakeness strategy.
+ * Location estimation occurs each round and are based on both previous estimations and newly received information.
+ */
+public class TracingSimulation extends Simulation {
+
+    private TracingSimulation(TracingSimulationBuilder builder) {
```
OrpazFishel, line 15: Once again I'll quote Alon:
"Prefer to move private methods (and constructors) below public methods. This way a reader of the class first sees the API they can use, and only then the implementation details".

src/test/java/com/google/research/bleth/simulator/TracingSimulationTest.java
```diff
@@ -0,0 +1,374 @@
...
+    private static final int BOARD_DIMENSION_EQUALS_THREE = 3;
+    private static final int MAX_ROUNDS_EQUALS_TWO = 2;
+    private static final int NUMBER_OF_BEACONS_EQUALS_TWO = 2;
+    private static final int NUMBER_OF_OBSERVERS_EQUALS_TWO = 2;
+    private static final double RADIUS_EQUALS_ONE = 1.0;
+
+    private static final int ILLEGAL_BOARD_DIMENSION = 0;
+    private static final int ILLEGAL_NUMBER_OF_AGENTS = 0;
+    private static final int ILLEGAL_MAX_ROUNDS = 0;
+    private static final int ILLEGAL_RADIUS = -1;
```
OrpazFishel, line 32: Nit: Consider name it NEGATIVE_RADIUS; I assume 0 is an illegal value for radius as well.

src/test/java/com/google/research/bleth/simulator/TracingSimulationTest.java
```diff
@@ -0,0 +1,374 @@
...
+        public Location moveTo(Board board, Location currentLocation) {
+            Location newLocation = currentLocation.moveInDirection(Direction.UP);
+            if (board.isLocationValid(newLocation)) {
+                return newLocation;
+            }
+            return currentLocation;
+        }
+    }
+
+    private static class FakeResolver implements IResolver {
```
OrpazFishel, line 363: And again:
"Consider splitting this to a separate file (even if it's not going to be reused - just so it doesn't clutter the tests file)".
(Unless you're waiting for the FakeResolver I wrote.)

src/main/java/com/google/research/bleth/simulator/GlobalResolver.java
```diff
@@ -0,0 +1,23 @@
+package com.google.research.bleth.simulator;
+
+import java.util.List;
+
+public class GlobalResolver implements IGlobalResolver {
```
nolaviz, line 5: Please add JavaDoc.

src/main/java/com/google/research/bleth/simulator/IGlobalResolver.java
```diff
@@ -0,0 +1,4 @@
+package com.google.research.bleth.simulator;
+
+public interface IGlobalResolver extends IResolver {
```
nolaviz, line 3: JavaDoc.

src/main/java/com/google/research/bleth/simulator/IResolver.java
```diff
@@ -0,0 +1,19 @@
+package com.google.research.bleth.simulator;
+
+import java.util.List;
+
+public interface IResolver {
```
nolaviz, line 5: JavaDoc.

src/main/java/com/google/research/bleth/simulator/Observer.java
```diff
@@ -0,0 +1,28 @@
+package com.google.research.bleth.simulator;
+
+public class Observer implements IObserver {
```
nolaviz, line 3: Guess what?
(JavaDoc)

src/main/java/com/google/research/bleth/simulator/TracingSimulation.java
nolaviz: +1.

src/main/java/com/google/research/bleth/simulator/TracingSimulation.java
```diff
@@ -0,0 +1,101 @@
...
+ * Observers change their awakeness states according to an awakeness strategy.
+ * Location estimation occurs each round and are based on both previous estimations and newly received information.
+ */
+public class TracingSimulation extends Simulation {
+
+    @Override
+    void updateSimulationStats() { }
+
+    private TracingSimulation(TracingSimulationBuilder builder) {
+
```
nolaviz, line 19: Nit: no need for a blank line here (or on line 25 as well).

src/main/java/com/google/research/bleth/simulator/TracingSimulation.java
```diff
@@ -0,0 +1,101 @@
...
+        super(builder.id, builder.currentRound, builder.maxNumberOfRounds, builder.resolver, builder.radius,
```
nolaviz, line 20: Instead of passing so many arguments to the `Simulation` constructor, prefer to just pass in the builder and have the constructor get values from it.

(Builders and constructors are allowed to be "extra friendly"... And passing more than 3-4 parameters to a method makes the code very hard to read, and easy to break accidentally).

src/main/java/com/google/research/bleth/simulator/Simulation.java
```diff
@@ -1,9 +1,399 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * An abstract class representing a BLETH simulation.
+ * Can be either a Tracing simulation or a Stalking simulation.
+ */
 public abstract class Simulation {
```
nolaviz, line 13: As this is a pure abstract class, consider renaming to AbstractSimulation for clarity.

src/main/java/com/google/research/bleth/simulator/Simulation.java
```diff
@@ -1,9 +1,399 @@
...
+    /**
+     * Write final simulation statistical data to db.
+     */
+    void writeSimulationStats() { }
+
+    /**
+     * An abstract builder class designed to separate the construction of a simulation from its representation.
+     * Designated for the construction of a new simulation, based on parameters given by an end-user.
+     */
+    public static abstract class SimulationBuilder {
```
nolaviz, line 112: Nit: since this is an inner class, common practice is to just call it "Builder" (and then always refer to it as "Simulation.Builder").

src/main/java/com/google/research/bleth/simulator/Simulation.java
```diff
@@ -1,9 +1,399 @@
...
+         * @return a new Simulation object constructed with the builder parameters.
+         */
+        public abstract Simulation build();
+    }
+
+    /**
+     * An abstract builder class designed to separate the construction of a simulation from its representation.
+     * Designated for the construction of a simulation based on old simulation state read from the db.
+     */
+    public static abstract class SimulationBuilderFromExisting {
```
nolaviz, line 260: Do you really need a separate builder class for this case? Can't you just add a factory method to `SimulationBuilder` to create it from existing state?

omermadmon: I can think of having a single `Builder` inner class, with both `buildFromScratch()` and `buildFromExisting()` methods. Did I get you right?

nolaviz:
Yes

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -0,0 +1,369 @@
...
+    /**
+     * Pass transmissions from beacons to observers while taking into consideration world-physics parameters,
+     * such as probability of transmission and distance between beacons and observers.
+     */
+    void beaconsToObservers() {
+        for (Beacon beacon : beacons) {
+            for (Observer observer : observers) {
+                double distance = distance(beacon.getLocation(), observer.getLocation());
+                if (distance <= radius) {
+                    observer.observe(beacon.transmit());
```
OrpazFishel, line 84: What if the observer isn't awake? You should check it first, maybe even before calculating the distance.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -0,0 +1,369 @@
...
+        }
+    }
+
+    /**
+     * Pass transmissions from beacons to observers while taking into consideration world-physics parameters,
+     * such as probability of transmission and distance between beacons and observers.
+     */
+    void beaconsToObservers() {
+        for (Beacon beacon : beacons) {
+            for (Observer observer : observers) {
```
OrpazFishel, line 81: Consider check the agents around the beacon instead of all the observers on the board.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -0,0 +1,369 @@
...
+        this.id = builder.id;
+        this.maxNumberOfRounds = builder.maxNumberOfRounds;
+        this.board = builder.realBoard;
+        this.resolver = builder.resolver;
+        this.radius = builder.radius;
+        this.beacons = ImmutableList.copyOf(builder.beacons);
+        this.observers = ImmutableList.copyOf(builder.observers);
+    }
+
+    private double distance(Location location1, Location location2) {
```
OrpazFishel, line 366: Consider name the parameters firstLocation and secondLocation.

src/main/java/com/google/research/bleth/simulator/AwakenessStrategyFactory.java
```diff
@@ -0,0 +1,48 @@
...
+     * @param awakenessDuration is the duration, which is the number of rounds in which an observer
+     * is awake in a single awakeness cycle.
+     * @return an awakeness strategy initialized with all required parameters.
+     */
+    public IAwakenessStrategy createStrategy(int awakenessCycle, int awakenessDuration) {
+        switch (this.type) {
+            case FIXED: return createFixedStrategy(awakenessCycle, awakenessDuration);
+            case RANDOM: return createRandomStrategy(awakenessCycle, awakenessDuration);
+        }
+        return null;
```
OrpazFishel, line 36: Consider using default instead, so the return value will never be null.

src/test/java/com/google/research/bleth/simulator/TracingSimulationIT.java
```diff
@@ -0,0 +1,63 @@
...
+                .setRealBoard(realBoard)
+                .setResolver(resolver)
+                .setBeacons(beacons)
+                .setObservers(observers)
+                .buildRestored();
+
+        simulation.run();
+
+        assertThat(beacon.getLocation()).isEqualTo(new Location(0, 0));
+        assertThat(observer.getLocation()).isEqualTo(new Location(0, 0));
```
OrpazFishel, line 61: Consider store this location and re-use it instead of create 3 identical instances.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -0,0 +1,369 @@
...
+    RealBoard getBoard() {
+        return board;
+    }
+
+    /**
+     * Run entire simulation logic, including writing data to db.
+     */
+    public void run() {
+        if (currentRound == 0) {
+            writeRoundState(); // round 0 is the initial simulation state
```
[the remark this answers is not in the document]
omermadmon: This will be more clear when implementing the `DatabaseService`, and we will probably indeed have a separate method for metadata writing.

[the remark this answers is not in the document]
omermadmon: Update: I actually think we want to write the simulation's metadata (and validating its ID uniqueness) during the `build()` method, even before its `run()` method is being called. If we do it only at the `run()` method we can end up having 2 simulations with the same ID in our db, and that's bad.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
omermadmon: Been thinking about it of course :)

This is more efficient and less readable, so we'll let readability win this tradeoff this time, and will optimize if performance is bad.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -0,0 +1,352 @@
...
+     * Pass transmissions from beacons to observers while taking into consideration world-physics parameters,
+     * such as probability of transmission and distance between beacons and observers.
+     */
+    void beaconsToObservers() {
+        for (Beacon beacon : beacons) {
+            for (Observer observer : observers) {
+                if (observer.isAwake()) {
+                    double distance = distance(beacon.getLocation(), observer.getLocation());
+                    if (distance <= radius) {
+                        observer.observe(beacon.transmit());
```
OrpazFishel, line 76: Consider separating the transmission from the observation, i.e. ask the beacon to transmit once and then send the transmission to the relevant observers. It represents reality more accurately and the method transmit may contain more logic in the future.

src/test/java/com/google/research/bleth/simulator/TracingSimulationBuilderTest.java
```diff
@@ -0,0 +1,239 @@
...
+    private static final int NUMBER_OF_BEACONS_EQUALS_TWO = 2;
+    private static final int NUMBER_OF_OBSERVERS_EQUALS_TWO = 2;
+    private static final double RADIUS_EQUALS_ONE = 1.0;
+    private static final int AWAKENESS_CYCLE_EQUALS_TWO = 2;
+    private static final int AWAKENESS_DURATION_EQUALS_ONE = 1;
+    private static final int AWAKENESS_DURATION_EQUALS_THREE = 3;
+    private static final AwakenessStrategyFactory.Type FIXES_AWAKENESS_STRATEGY_TYPE =
+            AwakenessStrategyFactory.Type.FIXED;
+
+    private static final int ILLEGAL_BOARD_DIMENSION = 0;
```
nolaviz, line 26: If a constant is used only once, prefer to just inline it to the specific test that uses it. (You can still give it a name, e.g. as "`final int illegalBoardDimension = 0`"; it just ensures the definition is close to the use.)

[In general, prefer to create constants for test values only when (a) the constant is useful in multiple tests, and (b) nothing matters about the constant except what's in its name.]

src/test/java/com/google/research/bleth/simulator/TracingSimulationIT.java
```diff
@@ -0,0 +1,75 @@
+package com.google.research.bleth.simulator;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class TracingSimulationIT {
```
nolaviz, line 10: I assume "IT" means something... Integration test maybe?
Prefer to spell this out fully; acronyms are confusing.

omermadmon: By default, the Failsafe Plugin automatically includes all test classes which follow this naming convention:
https://maven.apache.org/surefire/maven-failsafe-plugin/examples/inclusion-exclusion.html

(It can be manually configured if needed, but I guess following the 'IT' convention is ok).

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -0,0 +1,278 @@
...
+         * @param observerMovementStrategy is the movement strategy for all observers.
+         * @return this, to provide chaining.
+         */
+        public Builder setObserverMovementStrategy(IMovementStrategy observerMovementStrategy) {
+            this.observerMovementStrategy = observerMovementStrategy;
+            return this;
+        }
+
+        /** Write simulation metadata to db. */
+        abstract void writeSimulationMetadata();
```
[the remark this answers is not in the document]
omermadmon: These method include logic (validation and objects creation) required for to the simulation construction, and are being called inside the `build()` method:

```java
        @Override
        public AbstractSimulation build() {

            // validation of the correctness of the arguments provided using the builder's setters.
            validateArguments(); 

            // board and resolver creation.
            this.realBoard = new RealBoard(this.rowNum, this.colNum);
            this.resolver = new GlobalResolver(this.rowNum, this.colNum);

            // agents creation.
            // (the builder was only given number of beacons / observers, and their strategies).
            initializeBeacons();
            initializeObservers();

            // db interaction, and unique simulation id assignment.
            writeSimulationMetadata();

            // simulation construction.
            return new TracingSimulation(this);
        }
```

The only method not implemented yet is the `writeSimulationMetadata()`, and it should be something like:
```java
        @Override
        public void writeSimulationMetadata() {
            // Write an entity to the db, 
            // and return the unique id the simulation entity was assigned by datastore.
            this.id = DatabaseService.getInstance().writeSimulationMetadata(this); 
        }
```

These methods' logic can be copied into the `build()` method, but I think this will be less readable.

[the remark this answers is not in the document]
omermadmon: Update - the alternative might be moving these methods' logic to the setters:

- `validateArguments()` **is indeed redundant** - this validation can happen inside the setter methods.
- `initializeObservers()` required multiple fields to be already set - `observersNum`, `oberverMovementStrategy` and `awakenessStrategyType`. Thus the observers creation **cannot happen inside any of these setter methods.**
- `initializeBeacons()` is **similar to the previous method**. 
In addition, this shall be an **abstract class** due to the fact that in the stalking simulation the created beacons are instances of the `SwappingBeacon` class.
- `writeSimulationMetadata()` required all fields to be already set. As mentioned above, its implementation is a single-line thus we might **inline it into the `build()` method.**

If you agree with those statements I'll modify these to remove the redundant methods. Please let me know if I am missing something.

@OrpazFishel

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -0,0 +1,278 @@
...
+    void writeRoundState() { }
+
+    /** Gather statistical data of the current round and update the aggregated simulation statistics based on all rounds. */
+    abstract void updateSimulationStats();
+
+    /** Write final simulation statistical data to db. */
+    void writeSimulationStats() { }
+
+    /** An abstract builder class designed to separate the construction of a simulation from its representation. */
+    public static abstract class Builder {
```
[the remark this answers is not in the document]
omermadmon: Both _FreeBuilder_ and _AutoValue_ generate a builder with setters matching to the class attributes, i.e if the `AbstractSimulation` class has a `board` attribute, the generated builder will have a `setBoard()` method.
In our case, we want the builder to have `setRowNum()` and `setColNum()` methods, and during the `build()` it will create the `board` and `resolver` to be passed to the `AbstractSimulation` constructor.

(Same for other attributes, such as `beacons` and `observers` - we want the builder to have setter for the number of agents to create and their strategies, and the builder will create the agents itself).

Is there a way of using FreeBuilder or AutoValue to fit these requirements?

@OrpazFishel

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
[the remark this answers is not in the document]
nolaviz: Final (cross-field) validation indeed must happen during `build()`. I'm fine with that.
The same goes for the complex initialization steps.

As for database accesses - I thought the plan is to move them to separate (adapter) classes? (OK if this is done later.)

src/test/java/com/google/research/bleth/simulator/TracingSimulationIT.java
nolaviz: OK - I wasn't aware of this plugin (as usual, inside Google we have different tools).

nolaviz (approved):

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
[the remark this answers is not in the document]
omermadmon: Will be done later - those adapter classes (`BoardState` and `SimulationMetadata`) can be reviewed at #8.

Thanks!

PR #7, opened 2020-10-15
Add Global Resolver

OrpazFishel:


src/main/java/com/google/research/bleth/simulator/GlobalResolver.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.common.collect.HashBiMap;
+import com.google.common.collect.ImmutableBiMap;
+import com.google.common.collect.Multimap;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** Tracing Simulation's Resolver, which receives information from its observers and estimate the beacons' locations according to it. */
+public final class GlobalResolver implements IGlobalResolver {
+    private EstimatedBoard estimatedBoard;
```
nolaviz, line 15: Can these fields be marked final?

src/main/java/com/google/research/bleth/simulator/GlobalResolver.java
```diff
@@ -0,0 +1,82 @@
...
+    private Map<Beacon, Location> beaconsToEstimatedLocations = new HashMap<>();
+
+    /**
+     * A wrapper method to create new global resolver for a tracing simulation.
+     * The new resolver has a board for storing the estimated beacons' locations.
+     * @param rowsNum is number of rows of the simulation's board.
+     * @param colsNum is number of columns of the simulation's board.
+     * @param beacons is a list of the simulation's beacons.
+     */
+    public static GlobalResolver createResolver(int rowsNum, int colsNum, List<Beacon> beacons) {
```
nolaviz, line 27: Nit: since this method is inside a "resolver" class, prefer to just call it "create" (part of DRY: no need to repeat the "resolver")

src/main/java/com/google/research/bleth/simulator/GlobalResolver.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.common.collect.Multimap;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** Tracing Simulation's Resolver, which receives information from its observers and estimate the beacons' locations according to it. */
+public final class GlobalResolver implements IGlobalResolver {
+    private EstimatedBoard estimatedBoard;
+    private Multimap<Transmission, Location> currentRoundTransmissions = ArrayListMultimap.create();
+    private Map<Transmission, Beacon> TransmissionsToBeacons;
```
nolaviz, line 17: Field name should start with lowercase letter.

src/main/java/com/google/research/bleth/simulator/GlobalResolver.java
```diff
@@ -0,0 +1,82 @@
...
+     * A wrapper method to create new global resolver for a tracing simulation.
+     * The new resolver has a board for storing the estimated beacons' locations.
+     * @param rowsNum is number of rows of the simulation's board.
+     * @param colsNum is number of columns of the simulation's board.
+     * @param beacons is a list of the simulation's beacons.
+     */
+    public static GlobalResolver createResolver(int rowsNum, int colsNum, List<Beacon> beacons) {
+        checkNotNull(beacons);
+        EstimatedBoard estimatedBoard = new EstimatedBoard(rowsNum, colsNum);
+        HashBiMap<Transmission, Beacon> TransmissionsToBeacons = HashBiMap.create();
```
nolaviz, line 30: Why use a `BiMap` here? It's only used as a `Map` in the constructor (assigned to a `Map` field).
Prefer to just use a regular `ImmutableMap` instead - you can build it as

```
beacons.stream().collect(toImmutableMap(Beacon::transmit, Function.identity()))
```

nolaviz, line 30: Nit: variable name should start with a lowercase letter.

src/test/java/com/google/research/bleth/simulator/GlobalResolverTest.java
```diff
@@ -0,0 +1,254 @@
...
+        resolver.estimate();
+
+        assertThat(resolver.getBoard().getAgentsOnLocation(TWO_ON_TWO_COORDINATE)).containsExactly(beacon);
+    }
+
+    private Beacon createRandomBeaconOnLocation(Location initialLocation, RealBoard realBoard) {
+        return BEACON_FACTORY.createBeacon(initialLocation, new RandomMovementStrategy(), realBoard);
+    }
+
+    private boolean isBoardEmpty(Board board, int rowsNum, int colsNum) {
```
nolaviz, line 244: Doesn't a Board provide methods for getting its size directly? (Wouldn't having such methods simplify all callers that currently need to supply the size with the Board whenever they pass it somewhere else?)

nolaviz, line 244: If this method is for testing only, it's a bit too limiting (can only test a single state) + produces a rather weird-looking test (`assertThat(isBoardEmpty(...)).isTrue()`). The error message you'd get on failure would also be rather uninformative.

Consider instead providing a method like `agentsOnBoard(Board)`, which would collect all agents from the board and return them as a `Map<Location, Agent>`; you'd then have assertions like `assertThat(agentsOnBoard(...)).isEmpty()`, or `assertThat(agentsOnBoard(...)).containsExactly(...)` etc. - it could simplify nearly all of your testcases.

src/test/java/com/google/research/bleth/simulator/GlobalResolverTest.java
```diff
@@ -0,0 +1,254 @@
...
+                                                                /* colsNum= */ 2,
+                                                                ImmutableList.of(beacon));
+
+        resolver.estimate();
+
+        assertThat(isBoardEmpty(resolver.getBoard(), 2, 2));
+    }
+
+    @Test
+    public void firstEstimateAfterReceivingZeroTransmissions_EstimatedBoardIsEmpty() {
```
nolaviz, line 45: Nit: the "outcome" part of the name (after the underscore) should start again with a lowercase letter.
(Applies to other tests as well)

src/test/java/com/google/research/bleth/simulator/GlobalResolverTest.java
omermadmon: This method might be useful in another context as well: when writing the state of a Board object to the db we must iterate over all cells and retrieve the agents locations', so they could be recorded and written to the db.

Having a `agentsOnBoard()` method in `AbstractBoard` class will make the implementation of the `writeBoardState()` method more readable.

src/test/java/com/google/research/bleth/simulator/GlobalResolverTest.java
```diff
@@ -21,211 +21,211 @@ public void estimatedBoardInitializedEmpty() {
...
+        GlobalResolver resolver = GlobalResolver.create(/* rowsNum= */ 2,
+                                                        /* colsNum= */ 2,
+                                                        ImmutableList.of(beacon));
 
         resolver.receiveInformation(ZERO_ON_ZERO_COORDINATE, ImmutableList.of(beacon.transmit()));
         resolver.estimate();
 
-        assertThat(resolver.getBoard().getAgentsOnLocation(ZERO_ON_ZERO_COORDINATE)).containsExactly(beacon);
-        assertThat(resolver.getBoard().getAgentsOnLocation(ONE_ON_ONE_COORDINATE)).isEmpty();
+        assertThat(resolver.getBoard().agentsOnBoard().get(ZERO_ON_ZERO_COORDINATE)).containsExactly(beacon);
```
nolaviz, line 69: Now that you have `agentsOnBoard`, you could simplify both assertions here to a single

```
assertThat(resolver.getBoard().agentsOnBoard()
    .containsExactly(ImmutableListMultimap.of(ZERO_ON_ZERO_COORDINATE, beacon));
```

nolaviz: (similar comment apply to the other test cases; many can be simplified as well)

nolaviz (approved):

PR #8, opened 2020-10-22
Database Service

omermadmon:
`DatabaseService` is a singleton class wrapping the application's datastore instance, provides a global point of access to any read/write operations.

`DatabaseService` provides API for:
- Writing simulation metadata by the builder responsible for the simulation construction (which returns the unique id assigned to the simulation, encoded as a string).
- Writing a board state to the DB (by writing entities under the schema of `[simulationId, round, agentId, agentType, rowNum, colNum]` ).
- Reading a board state from the DB (which returns null if provided round exceeds simulation's max number of round, or a JSON string encoding the board state).

In addition, `DatabaseService`  throw informative exceptions whenever an invalid read/write operation occurs (e.g when writing a board state which already exists for a certain `simulationId` and `round`).


src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,225 @@
...
+                new Query.FilterPredicate(Entity.KEY_RESERVED_PROPERTY, Query.FilterOperator.EQUAL, simulationKey);
+        Query q = new Query("Simulation").setFilter(filterBySimulationId);
+        PreparedQuery pq = datastore.prepare(q);
+        return pq.asSingleEntity();
+    }
+
+    private boolean isRoundExistsInSimulation(String simulationId, int round) {
+        Entity simulationEntity = retrieveSimulationEntity(simulationId);
+        int maxSimulationRound = ((Long) simulationEntity.getProperty("roundsNum")).intValue();
+        return round < maxSimulationRound;
```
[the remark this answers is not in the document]
omermadmon: I agree something needs to be changed, though I am not sure if we should change it here or at the Simulation's `run()` method (to be `currentRound < maxNumberOfRounds` as well).

It depends whether we consider the initial state (round zero) as a round, or not.

In my opinion it is more reasonable to count it as a round, otherwise we will have rounds 0 to maxRound recorded at out db - these are maxRound + 1 rounds instead of maxRound rounds (which is a bit weird, I think).

(In other words - I think we want the last round to be the maxRound - 1 round, thus we need to change the while loop condition of the simulation, and leave it as is here).

src/main/java/com/google/research/bleth/exceptions/ExceedingRoundException.java
```diff
@@ -0,0 +1,7 @@
+package com.google.research.bleth.exceptions;
+
+public class ExceedingRoundException extends RuntimeException {
+    public ExceedingRoundException(String message) {
+        super(message);
+    }
```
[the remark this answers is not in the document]
omermadmon: It is now my turn to quote Alon:

"Prefer to create your own exception classes, rather than inventing new semantics for existing exceptions."

src/main/java/com/google/research/bleth/services/DatabaseService.java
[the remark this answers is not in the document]
OrpazFishel: I assume that considering the initial state as round 0 will be more comfortable, but it's your call. Just make sure it's consistent.

src/main/java/com/google/research/bleth/exceptions/ExceedingRoundException.java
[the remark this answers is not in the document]
OrpazFishel: Oh I do remember this comment, I just don't see why this is new semantics. But again - it's your call.

src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,223 @@
...
+
+        // Create an empty table representing the board state.
+        ArrayTable<Integer, Integer, ArrayList<String>> boardState = createEmptyTable(rowNum, colNum);
+
+        // Fill board state with agents' string representations according to query result.
+        for (Entity entity : boardStateEntities) {
+            int row = ((Long) entity.getProperty("rowNum")).intValue();
+            int col = ((Long) entity.getProperty("colNum")).intValue();
+            String type = (String) entity.getProperty("agentType");
+            Long id = (Long) entity.getProperty("agentId");
```
OrpazFishel, line 177: Nit: Is there any reason for using Long instead of long?

src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,223 @@
...
+        // Set properties.
+        entity.setProperty("type", builder.getSimulationType());
+        entity.setProperty("roundsNum", builder.getMaxNumberOfRounds());
+        entity.setProperty("beaconsNum", builder.getBeaconsNum());
+        entity.setProperty("observersNum", builder.getObserversNum());
+        entity.setProperty("rowNum", builder.getRowNum());
+        entity.setProperty("colNum", builder.getColNum());
+        entity.setProperty("beaconMovementStrategy", builder.getBeaconMovementStrategy().toString());
+        entity.setProperty("observerMovementStrategy", builder.getObserverMovementStrategy().toString());
+        entity.setProperty("observerAwakenessStrategy", builder.getAwakenessStrategyType().toString());
```
OrpazFishel, line 70: Do all the observers have the same strategy? If so - it's observer**s** and not observer, otherwise we should save this information about each agent.

src/test/java/com/google/research/bleth/services/DatabaseServiceTest.java
```diff
@@ -0,0 +1,257 @@
...
+    @Before
+    public void mockAgents() {
+        Mockito.when(beacon.getId()).thenReturn(0);
+        Mockito.when(beacon.getType()).thenReturn("Beacon");
+        Mockito.when(observer.getId()).thenReturn(0);
+        Mockito.when(observer.getType()).thenReturn("Observer");
+    }
+
+    @Test
+    public void writeNonEmptyRealBoardThenReadRealBoard_shouldGetExpectedJson() {
```
OrpazFishel, line 110: Consider mentioning that you should get this Json when you read from db - the name only mentions writing to it.

src/test/java/com/google/research/bleth/services/DatabaseServiceTest.java
```diff
@@ -0,0 +1,257 @@
...
+        DatabaseService db = DatabaseService.getInstance();
+        RealBoard realBoard = new RealBoard(BOARD_DIMENSION, BOARD_DIMENSION);
+        realBoard.placeAgent(ZERO_ON_ZERO_COORDINATE, beacon);
+        realBoard.placeAgent(ONE_ON_ONE_COORDINATE, observer);
+        String firstSimulationId = db.writeMetadata(firstSimulationBuilder);
+
+        db.writeBoardState(firstSimulationId, ZERO_ROUND, realBoard);
+
+        assertThrows(BoardStateAlreadyExistsException.class, () -> {
+            db.writeBoardState(firstSimulationId, 0, realBoard);
```
OrpazFishel, line 221: Why don't you use here ZERO_ROUND?

src/main/java/com/google/research/bleth/services/DatabaseService.java
omermadmon: Yes, the `intValue()` method.

src/test/java/com/google/research/bleth/services/DatabaseServiceTest.java
omermadmon: writeNonEmptyRealBoard**ThenRead**RealBoard_shouldGetExpectedJson

src/main/java/com/google/research/bleth/services/DatabaseService.java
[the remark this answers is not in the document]
omermadmon: Resolved at `omermadmon-simulation` branch :)

src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,223 @@
...
+ * assigned to the simulation, encoded to a string).
+ * - Writing a board state to the db.
+ * - Reading a board state from the db (which returns null if provided round exceeds simulation's max number of round,
+ * or a json string encoding the board state).
+ */
+public class DatabaseService {
+
+    private static DatabaseService instance = null;
+    private DatastoreService datastore;
+    private Gson gson = new Gson();
```
schainmariano, line 39: purpose of this?

src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,223 @@
...
+        // Create new entity.
+        Entity entity = new Entity("Simulation");
+
+        // Set properties.
+        entity.setProperty("type", builder.getSimulationType());
+        entity.setProperty("roundsNum", builder.getMaxNumberOfRounds());
+        entity.setProperty("beaconsNum", builder.getBeaconsNum());
+        entity.setProperty("observersNum", builder.getObserversNum());
+        entity.setProperty("rowNum", builder.getRowNum());
+        entity.setProperty("colNum", builder.getColNum());
```
schainmariano, line 67: the database service seems to be aware of internal structure of the simulation object (e.g., why should it care about the movement strategy?). this creates a dependency in the wrong direction.

src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,223 @@
...
+
+        // Compose simple predicates.
+        Query.CompositeFilter composedFilter = Query.CompositeFilterOperator.and(filterBySimulationId, filterByRound);
+
+        // Create query and return result.
+        Query boardStateBySimulationIdAndRoundQuery = new Query(entityKind).setFilter(composedFilter);
+        PreparedQuery boardStateBySimulationIdAndRoundPreparedQuery = datastore.prepare(boardStateBySimulationIdAndRoundQuery);
+
+        // Return encoded board state.
+        return toJsonTable(boardStateBySimulationIdAndRoundPreparedQuery.asIterable(), simulationId);
```
schainmariano, line 160: can this call fail?

src/main/java/com/google/research/bleth/services/DatabaseService.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    private String toJsonTable(Iterable<Entity> boardStateEntities, String simulationId) {
+        // Retrieve simulations' dimensions.
+        Entity simulationEntity = retrieveSimulationEntity(simulationId);
+        int rowNum = ((Long) simulationEntity.getProperty("rowNum")).intValue();
```
schainmariano, line 166: this highly couples the database utility with the simulation implementation/interface

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -0,0 +1,33 @@
...
+
+/** A placeholder class for representing a BLETH simulation. */
+public abstract class AbstractSimulation {
+
+    /** A placeholder class for representing a BLETH simulation builder. */
+    public static abstract class Builder {
+
+        public abstract String getSimulationType();
+
+        public int getMaxNumberOfRounds() { return 1; }
```
schainmariano, line 11: if this is an abstract class then you should not implement pure abstract methods (that is, do not "return 1")

schainmariano (changes requested):
my main concern is the dependency of the service on the internals of the client

src/main/java/com/google/research/bleth/services/DatabaseService.java
omermadmon: The idea here was that the `builder` will pass himself to the `writeSimulationMetadata()` method inside its `build()` method. Something like:

```java
public TracingSimulation build() {
    this.id = DatabaseService.getInstance().writeSimulationMetadata(this);
    return new TracingSimulation(this);
}
```

Alternatively we can modify `writeSimulationMetadata()` to recieve multiple arguments instead of a builder, and then our `build()` method will look like:

```java
public TracingSimulation build() {
    this.id = DatabaseService.getInstance().writeSimulationMetadata(this.rowNum, this.colNum, /** lots of other attributes */);
    return new TracingSimulation(this);
}
```

But this smells bad as well. I guess what we're talking about is having some other object in the middle, which specifies both the properties names and the values to be written to the db. Let's say we're using some kind of a HashMap:

```java
public TracingSimulation build() {
    HashMap<String, Double> metadataEntity = new HashMap<>();
    metadataEntity.put("rowNum", this.rowNum);
    metadataEntity.put("colNum", this.colNum);
    /** lots of other attributes */
    this.id = DatabaseService.getInstance().write(metadataEntity);
    return new TracingSimulation(this);
}
```

And this might be better, because now the `write()` method can be similar for all entities we're writing. But someone must create this HashMap (or whatever we will work with) everytime, including specifing the properties names. 

Did you think of something else? @schainmariano

omermadmon: Also, the reason we're writing the Simulation's strategies as part of the metadata is to able to provide answers for questions like "what was the average result in all simulations where the beacons moved **randomally** and the observers where **static**?"

src/main/java/com/google/research/bleth/services/DatabaseService.java
omermadmon: Here the database utility is dependent on the datastore Simulation entity structure (i.e its schema), not the simulation implementation/interface. How can we avoid this? when reading an entity from datastore (or any other relational/non-relational db) we must specify the entity's properties names.

@schainmariano

src/main/java/com/google/research/bleth/services/DatabaseService.java
omermadmon: - `asIterable()` throws an exception if the query being executed is associated with a `Transaction` that is not active:
https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/PreparedQuery.html#asIterable--
This doesn't seem to be the case here since we're not using Datastore Transactions.
- `toJsonTable()` throws an exception if the simulation associated with the specified `simulationId` does not exist in datastore. But if that is the case, an exception will be thrown even before that, right when there is a call for `isRoundExistsInSimulation(simulationId, round)` (line 141).

These are all unchecked exception, therefore there is no use of the `throws` keyword in the method's signature. I will add some documentation to explain this behaviour.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
omermadmon: This `AbstractSimulation` is a placeholder, and it exists only to provide a mock simulation in tests. This implementation will be overridden as soon as PR #6 is merged to master.

omermadmon:
We are aiming for a DB service which is more flexible and agnostic, hence we can just use the Datastore API without writing a wrapper class.

In addition, I have added the following classes: 
- `BoardState` -  represents a static `Board` associated with a specific `simulationId` and `round`.
Provides API for writing to db, (statically) reading from db and JSON serialization.
- `SimulationMetadata` - stores all metadata associated with a simulation.
Provides API for writing to db and (statically) reading from db.
- `Schema` - stores an provides a single access point to all Datastore entities kinds and schemas.

This way our database service does not depend on its clients, and also only the `BoardState` and `SimulationMetadata` classes depend on the database service.
i.e: if we decide to use Cloud SQL instead of Datastore, we know we will have to change only those classes.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -0,0 +1,38 @@
...
+
+        public AwakenessStrategyFactory.Type getAwakenessStrategyType() { return null; }
+
+        public double getRadius() { return 1; }
+
+        public int getAwakenessCycle() { return 1; }
+
+        public int getAwakenessDuration() { return 1; }
+
+        /** Create an entity storing the simulation's metadata and store it on db. */
```
OrpazFishel, line 33: Nit: Consider re-write the sentence.

src/main/java/com/google/research/bleth/simulator/BoardState.java
```diff
@@ -0,0 +1,176 @@
...
+     * @param rows is the number of rows.
+     * @param cols is the number of columns.
+     * @param matrix is an array table of array lists of strings, where each string encodes the agent's type and id.
+     * @param simulationId is the associated simulation id.
+     * @param round is the associated round.
+     * @param entityKind is a string indicating the entity kind to create when this BoardState is written to datastore.
+     *                   can be either "RealBoardState" or "EstimatedBoardState".
+     */
+    BoardState(int rows, int cols, ArrayTable<Integer, Integer, ArrayList<String>> matrix,
+                      String simulationId, int round, String entityKind) {
```
OrpazFishel, line 42: Nit: Indentation.

src/main/java/com/google/research/bleth/simulator/BoardState.java
```diff
@@ -0,0 +1,176 @@
...
+                Objects.equal(entityKind, that.entityKind);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(matrix, simulationId, round, entityKind);
+    }
+
+    /**
+     * Create multiple datastore entities (each represents and agent in its location) associated with the board state,
```
OrpazFishel, line 68: Nit: Did you mean "represents an agent"?

src/main/java/com/google/research/bleth/simulator/BoardState.java
```diff
@@ -0,0 +1,176 @@
...
+     * and write them to the db.
+     * @throws BoardStateAlreadyExistsException if a board state associated with the same simulation id, round and entity kind
+     * already exists in the db.
+     */
+    public void write() throws BoardStateAlreadyExistsException {
+        // Throw and exception if a board state associated with the same simulation id, round and entity kind
```
OrpazFishel, line 74: Nit:  Did you mean "an exception"?

src/main/java/com/google/research/bleth/simulator/BoardStateFactory.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Create an empty array table of array lists of strings.
+     * @param rowNum is the number of rows.
+     * @param colNum is the number of columns.
+     * @return an empty array table of array lists of strings.
+     */
+    static ArrayTable<Integer, Integer, ArrayList<String>> createEmptyTable(int rowNum, int colNum) {
+        ArrayTable<Integer, Integer, ArrayList<String>> table;
+        table = ArrayTable.create(IntStream.range(0, rowNum).boxed().collect(Collectors.toList()),
+                IntStream.range(0, colNum).boxed().collect(Collectors.toList()));
```
OrpazFishel, line 42: Nit: Indentation.

src/main/java/com/google/research/bleth/simulator/BoardStateFactory.java
```diff
@@ -0,0 +1,71 @@
...
+                boardStateTable.get(location.row, location.col).add(agent.getType() + agent.getId());
+            }
+        }
+
+        return boardStateTable;
+    }
+
+    private static String determineBoardStateEntityKind(Board board) {
+        if (board instanceof RealBoard) { return Schema.BoardState.entityKindReal; }
+        return Schema.BoardState.entityKindEstimated;
```
OrpazFishel, line 69: Make sure the board isn't null.

src/main/java/com/google/research/bleth/simulator/SimulationMetadata.java
```diff
@@ -0,0 +1,114 @@
...
+        this.observerMovementStrategy = builder.getObserverMovementStrategy().toString();
+        this.observerAwakenessStrategy = builder.getAwakenessStrategyType().toString();
+        this.radius = builder.getRadius();
+        this.awakenessCycle = builder.getAwakenessCycle();
+        this.awakenessDuration = builder.getAwakenessDuration();
+        this.observersDensity = (double) builder.getObserversNum() / (builder.getRowNum() * builder.getColNum());
+        this.awakenessRatio = (double) builder.getAwakenessDuration() / builder.getAwakenessCycle();
+    }
+
+    /** Create an entity storing the simulation's metadata and store it on db. */
```
OrpazFishel, line 49: Nit: Consider re-write the sentence.

src/main/java/com/google/research/bleth/simulator/SimulationMetadata.java
```diff
@@ -0,0 +1,114 @@
...
+        entity.setProperty(Schema.SimulationMetadata.beaconMovementStrategy, this.beaconMovementStrategy);
+        entity.setProperty(Schema.SimulationMetadata.observerMovementStrategy, this.observerMovementStrategy);
+        entity.setProperty(Schema.SimulationMetadata.observerAwakenessStrategy, this.observerAwakenessStrategy);
+        entity.setProperty(Schema.SimulationMetadata.radius, this.radius);
+        entity.setProperty(Schema.SimulationMetadata.awakenessCycle, this.awakenessCycle);
+        entity.setProperty(Schema.SimulationMetadata.awakenessDuration, this.awakenessDuration);
+        entity.setProperty(Schema.SimulationMetadata.observersDensity, this.observersDensity);
+        entity.setProperty(Schema.SimulationMetadata.awakenessRatio, this.awakenessRatio);
+
+        // Write to datastore and return key (as string).
```
OrpazFishel, line 70: The parentheses aren't necessary.

src/main/java/com/google/research/bleth/simulator/SimulationMetadata.java
```diff
@@ -0,0 +1,114 @@
...
+        Query.Filter filterBySimulationId =
+                new Query.FilterPredicate(Entity.KEY_RESERVED_PROPERTY, Query.FilterOperator.EQUAL, simulationKey);
+        Query simulationIdQuery = new Query(Schema.SimulationMetadata.entityKind).setFilter(filterBySimulationId);
+        PreparedQuery simulationIdPreparedQuery = datastore.prepare(simulationIdQuery);
+        Entity simulationMetadataEntity = simulationIdPreparedQuery.asSingleEntity();
+        return new SimulationMetadata(simulationMetadataEntity);
+    }
+
+    /** Return true if provided round exists in the simulation associated with the provided simulation id, and false otherwise. */
+    static boolean roundExistsInSimulation(String simulationId, int round) {
```
OrpazFishel, line 92: Consider renaming to isRoundExistsInSimulation.

schainmariano (approved):

PR #9, opened 2020-10-25
Change Location and Transmission to AutoValue's classes

OrpazFishel:


src/main/java/com/google/research/bleth/simulator/Location.java
```diff
@@ -1,40 +1,23 @@
...
+    public abstract int col();
 
     /**
      * Calculate and return the new location based on the current location and the given direction.
      * @param direction is the direction to move to it, one of the four directions: up, down, right and left.
      * @return new Location, differ from the original location in one step, according the given direction.
      */
     public Location moveInDirection(Direction direction) {
-        return new Location(row + direction.getRowDelta(), col + direction.getColDelta());
+        return Location.create(row() + direction.getRowDelta(), col() + direction.getColDelta());
```
nolaviz, line 21: Nit: no need for `Location.create` here (you can just call the method `create`, as it's in the current scope).

nolaviz (approved):

PR #10, opened 2020-11-03
Add integration tests for tracing simulation

OrpazFishel:


src/test/java/com/google/research/bleth/simulator/TracingSimulationIT.java
```diff
@@ -13,65 +16,159 @@
...
+                .setTransmissionThresholdRadius(transmissionRadius)
+                .setBeaconMovementStrategy(MOVE_UP)
+                .setObserverMovementStrategy(STATIONARY)
+                .setAwakenessCycle(AWAKENESS_CYCLE_EQUALS_TWO)
+                .setAwakenessDuration(AWAKENESS_DURATION_EQUALS_ONE)
+                .setAwakenessStrategyType(FIXED_AWAKENESS_STRATEGY_TYPE)
+                .build();
+
+        Map<IAgent, Location> initialAgentsToLocations = mapAgentsToLocationsOnBoard(simulation.getBoard());
+        Set<IAgent> observers = initialAgentsToLocations.keySet().stream().filter(agent -> agent instanceof Observer).collect(Collectors.toSet());
```
omermadmon, line 113: Why not just 
```java 
Set<IAgent> observers = simulation.observers.stream().collect(Collectors.toSet()); 
```
?

src/test/java/com/google/research/bleth/simulator/TracingSimulationIT.java
```diff
@@ -13,65 +16,159 @@
...
+        Location observerInitialLocation = averageLocationOfObservers(observers);
```
omermadmon, line 114: We have a single observer here, why do we compute the average location?
(If the reason is to fit the pattern of the other tests, that's fine, just wanted to make sure).

omermadmon (commented):
Looks good to me.
In general, we want to have these four levels of test classes:
- Builder tests - to verify the builders' verification logic works properly in edge-cases (already exists).
- Database-only tests - involving those classes which interact with the database, but not involving the `Simulation` class (these tests exist at #8).
- Simulation-only tests - involving the `Simulation` class, but not involving the db-related classes (these are the tests you have just wrote).
- End-to-end tests - involving both the simulation and the database interaction. Those tests will be written as soon as the db-related classes are merged to the master.

src/test/java/com/google/research/bleth/simulator/TracingSimulationIT.java
OrpazFishel: I will change if simulation.observers remains protected and not private.
(Then simulation.observers will be good enough)

OrpazFishel:
Refactor using BoardState.

PR #11, opened 2020-11-03
Strategies Mapper

omermadmon:
When writing a simulation metadata to the db, a string representing the movement/awakeness strategy type is being written as a property.
The `StrategiesMapper` class provides API for:
- Generating a list of strings representing all  existing movement/awakeness strategies. 
When the end user will fill the form for new simulation creation, he will be able to choose values from this closed list.
- Retrieving a `AwakenessStrategyFactory.Type` or `MovementStrategyFactory.Type` based on the string representing the relevant strategy type.
When a new simulation is being created, the `StrategiesMapper` is used to generate this type, and then this type is being set to the `SimulationBuilder`.


src/main/java/com/google/research/bleth/simulator/StrategiesMapper.java
```diff
@@ -0,0 +1,76 @@
...
+     * Create and return a new movement strategy based on the type provided as a string.
+     * @param typeAsString a string indicating the type of movement strategy to create.
+     * @return a movement strategy object.
+     * @throws NoSuchMethodException if the concrete movement strategy does not have a declared constructor.
+     * @throws IllegalAccessException if the concrete movement strategy's constructor is not accessible.
+     * @throws InstantiationException if the instantiation fails.
+     */
+    public IMovementStrategy getMovementStrategy(String typeAsString)
+            throws NoSuchMethodException, IllegalAccessException, InvocationTargetException, InstantiationException {
+        return movement.get(typeAsString).getDeclaredConstructor().newInstance();
```
OrpazFishel, line 43: Works for now, won't work when we add the pre-defined strategy.

src/main/java/com/google/research/bleth/simulator/SimulationMetadata.java
```diff
@@ -36,8 +36,8 @@ public SimulationMetadata(AbstractSimulation.Builder builder) {
         this.observersNum = builder.getObserversNum();
         this.rowsNum = builder.getRowNum();
         this.colsNum = builder.getColNum();
-        this.beaconMovementStrategy = builder.getBeaconMovementStrategy().toString();
-        this.observerMovementStrategy = builder.getObserverMovementStrategy().toString();
+        this.beaconMovementStrategy = builder.getBeaconMovementStrategy().getClass().toString();
```
OrpazFishel, line 39: Does the builder hold a specific instance - i.e. the movement strategy of a specific agent?
If we can create new instance of the strategy from a string - why not store that string?

omermadmon: Actually that's a good point. Right now the Simulation Builder gets an instance of a movement strategy, and initialize all beacons/observers with that instance. We actually want each agent have its own movement strategy object (even if they are all instances of the same class). 

To solve this we can use the same solution we used to create awakeness strategies (the builder gets the strategy type, and the strategies instances created by a factory).

src/main/java/com/google/research/bleth/simulator/StrategiesMapper.java
omermadmon: I agree. Will be refactored as soon as we will have such a strategy.

src/main/java/com/google/research/bleth/simulator/SimulationMetadata.java
OrpazFishel: Sounds good.

src/main/java/com/google/research/bleth/simulator/BoardStateFactory.java
```diff
@@ -57,7 +57,7 @@ public static BoardState create(Board board, String simulationId, int round) {
         int colNum = board.getColNum();
         ArrayTable<Integer, Integer, ArrayList<String>> boardStateTable = createEmptyTable(rowNum, colNum);
         Multimap<Location, IAgent> agentsOnBoard = board.agentsOnBoard();
-        for (Location location : agentsOnBoard.keys()) {
+        for (Location location : agentsOnBoard.keySet()) {
```
nolaviz, line 60: Nit: this seems like an independent fix.
Prefer to fix such issues in a separate pull-request, to make review easier.

nolaviz (approved):

PR #13, opened 2020-11-06
New Simulation Form

omermadmon:
- `ListStrategiesServlet` - A servlet used for displaying all string representations of movement and awakeness strategies.
- `new_simulation.html` - An HTML page with a form for initializing a new simulation (no `submit` button or `NewSimulationServlet` yet).
- `new_simulation.js` - A javascript file providing functions for: 
a. Fetching `ListStrategiesServlet` and enabling choosing strategies from a closed list.
b. Setting dynamic upper bounds, such as: `awakenessCycle <= roundsNum` and `awakenessDuration <= awakenessCycle`.

image

### DIFF
--- a/src/main/java/com/google/research/bleth/servlets/ListStrategiesServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/ListStrategiesServlet.java
@@ -1,7 +1,7 @@
 package com.google.research.bleth.servlets;
 
-import com.google.appengine.repackaged.com.google.gson.Gson;
-import com.google.appengine.repackaged.com.google.gson.JsonArray;
+import com.google.gson.Gson;
+import com.google.gson.JsonArray;
 import com.google.research.bleth.simulator.StrategiesMapper;
 
 import java.io.IOException;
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/research/bleth/servlets/ListStrategiesServlet.java
```diff
@@ -0,0 +1,33 @@
+package com.google.research.bleth.servlets;
+
+import com.google.appengine.repackaged.com.google.gson.Gson;
```
OrpazFishel, line 3: What is the difference between this and com.google.gson.Gson?

omermadmon: The two methods of importing Gson require different maven dependencies. We have used the other method on the branch which didn't end up merged to the master branch (`omermadmon-gson-datastore`), that's where the confusion comes from.

Also, it seems like com.google.appengine.repackaged.com.google.gson.Gson is less recommended, so I will modify it. Thanks.

OrpazFishel (approved):

PR #14, opened 2020-11-07
Read Board State Servlet

omermadmon:
Added a servlet used for reading a board state from the db and write it to response as a JSON 2D array.

### DIFF
--- a/src/main/java/com/google/research/bleth/servlets/ReadBoardStateServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/ReadBoardStateServlet.java
@@ -1,0 +1,35 @@
+package com.google.research.bleth.servlets;
+
+import com.google.research.bleth.simulator.BoardState;
+
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** A servlet used for reading a board state from the db and write it to response as a JSON 2D array. */
+@WebServlet("/read-board-state")
+public class ReadBoardStateServlet extends HttpServlet {
+
+    @Override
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/research/bleth/servlets/ReadBoardStateServlet.java
```diff
@@ -0,0 +1,35 @@
...
+
+        // Get board state parameters.
+        int round = Integer.parseInt(request.getParameter("round"));
+        String simulationId = request.getParameter("simulationId");
+        boolean isReal = Boolean.parseBoolean(request.getParameter("isReal"));
+
+        // Read board state.
+        BoardState boardState;
+        if (isReal) {
+            boardState = BoardState.readReal(simulationId, round);
```
schainmariano, line 26: I see that the BoardState serves to retrieve boardState instances given simulationId. Why two different functions for real/non-real?

schainmariano (commented):
just a bit of clarification needed wrt isReal

src/main/java/com/google/research/bleth/servlets/ReadBoardStateServlet.java
omermadmon: The combination of `simulationId` and `round` defines a single real board state and a single estimated board state. Thus when reading a board state we need to somehow specify the wanted kind.
`isReal` is a property of the HTTP GET request that determines which read method should be called. An alternative would be having a single `BoardState.read()` method which gets `isReal` as a third argument, but as @nolaviz have mentioned:

"Google best practices recommend to avoid boolean parameters as much as possible; instead, we recommend to use two separate methods, whose names reflect the effect of the boolean, and possibly sharing aspects of the implementation."
https://github.com/OrpazFishel/step/pull/7

(Another alternaitive would be having two seperate servlets, one for reading a real board state and one for estimated; but that seems like an overkill..)

schainmariano:
What is the difference between real and estimated. If estimated, by whom?

schainmariano (approved):

PR #16, opened 2020-11-09
Gather statistical data during the simulation and write it to Datastore at the end

OrpazFishel:


src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -51,6 +52,11 @@ public void run() {
         writeSimulationStats();
     }
 
+    /** Returns the simulation's Id. */
+    public String getId() {
+        return id;
+    }
```
omermadmon, line 58: Why should a `Simulation` object return its `simultaionId`? Is it for testing purposes?

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -68,15 +74,21 @@ void updateObserversAwaknessState() {
...
                     double distance = distance(beacon.getLocation(), observer.getLocation());
                     if (distance <= transmissionThresholdRadius) {
                         observer.observe(transmission);
+                        observed = true; // for stats
                     }
                 }
             }
+            if (observed) {
+                beaconsObservedPercent.put(String.valueOf(beacon.getId()),
+                        beaconsObservedPercent.getOrDefault(String.valueOf(beacon.getId()), 0D) + 1); // for stats
```
omermadmon, line 90: `beaconsObservedPercent` actually holds the observations count, not percent. Consider changing its name or alternatively divide by number of rounds each iteration.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -94,10 +106,35 @@ void resolverEstimate() {
     void writeRoundState() { }
 
     /** Gather statistical data of the current round and update the aggregated simulation statistics based on all rounds. */
-    void updateSimulationStats() { }
+    void updateSimulationStats() {
```
omermadmon, line 109: I would suggest calculating the empirical standard deviation of the distances as well.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -94,10 +106,35 @@ void resolverEstimate() {
...
+        if (!distances.isEmpty()) { // distances is empty until the first round an observer observed a beacon
+            double min = distances.stream().min(Double::compareTo).get();
+            distancesStats.put("min", Math.min(distancesStats.getOrDefault("min", Double.POSITIVE_INFINITY), min));
+
+            double max = distances.stream().max(Double::compareTo).get();
+            distancesStats.put("max", Math.max(distancesStats.getOrDefault("max", Double.NEGATIVE_INFINITY), max));
+
+            double average = distances.stream().mapToDouble(Double::doubleValue).sum() / distances.size();
+            double allRoundsAverage = (distancesStats.getOrDefault("avg", 0D) * (currentRound - 1) + average) / currentRound;
+            distancesStats.put("avg", allRoundsAverage);
```
omermadmon, line 123: The problem with `average` (and `allRoundsAverage`) is the fact that it is not effected by the number of beacons estimated by the resolver (but only by the estimations quallity). Few suggestions:

- For each unobserved beacon, set its distance as the maximum of the known distances (or any other manipulation that will increase the average for every unobserved beacon).
- Multiply the pure average by a factor which is growing as the number of unobserved beacons is growing, e.g: 
`# num of unobserved beacons` or `# num of unobserved beacons / # total num of beacons`

src/main/java/com/google/research/bleth/simulator/Schema.java
```diff
@@ -31,4 +31,15 @@
         public static final String rowNum = "rowNum";
         public static final String colNum = "colNum";
     }
+
+    /** A class providing a single access point for the StatsState entity schema. */
+    public static class StatisticsState {
```
omermadmon, line 36: Can we store the names of the statistical measures (i.e max, min, avg, etc) as a static class of `Schema` as well?

src/main/java/com/google/research/bleth/simulator/StatisticsState.java
```diff
@@ -0,0 +1,109 @@
...
+     * Create and write multiple datastore entities, each represents an aggregate function such as min and max,
+     * of the distance between the beacons' real locations and their estimated locations.
+     */
+    public void writeDistancesStats() {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        for (Map.Entry<String, Double> entry : distanceStats.entrySet()) {
+            Entity entity = new Entity(Schema.StatisticsState.entityKindDistance);
+            entity.setProperty(Schema.StatisticsState.simulationId, simulationId);
+            entity.setProperty(Schema.StatisticsState.aggregateFunction, entry.getKey());
+            entity.setProperty(Schema.StatisticsState.value, entry.getValue());
```
omermadmon, line 42: This stores a new entity for each measure (i.e an entity for 'max', another entity for 'min' property, etc).
I think we should store a single entity with all measure as properties. This will be easier to query and also saves storage).

src/main/java/com/google/research/bleth/simulator/StatisticsState.java
```diff
@@ -0,0 +1,109 @@
...
+     * has been observed by at least one observer, i.e. has been detected by the resolver.
+     */
+    public void writeBeaconsObservedPercentStats() {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        for (String beaconId : beaconsObservedPercent.keySet()) {
+            Entity entity = new Entity(Schema.StatisticsState.entityKindBeaconsObservedPercent);
+            entity.setProperty(Schema.StatisticsState.simulationId, simulationId);
+            entity.setProperty(Schema.StatisticsState.beaconId, beaconId);
+            entity.setProperty(Schema.StatisticsState.percent, beaconsObservedPercent.get(beaconId));
+            datastore.put(entity);
```
omermadmon, line 58: Do we really want to store this measure for each beacon? 
Even if we do, I would also store the average over all beacons (it is easier to calculate it here rather than by querying the db).

src/main/java/com/google/research/bleth/simulator/StatisticsState.java
```diff
@@ -0,0 +1,109 @@
...
+     * @param simulationId is the simulation id associated with the statistical data.
+     * @return a map that maps a beacon's id to the percentage of rounds a beacon has been observed by at least one observer.
+     */
+    public static Map<String, Double> readBeaconsObservedPercentStats(String simulationId) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        ImmutableMap.Builder<String, Double> beaconsObserved = new ImmutableMap.Builder<>();
+
+        Query.FilterPredicate filterBySimulationId =
+                new Query.FilterPredicate(Schema.StatisticsState.simulationId, Query.FilterOperator.EQUAL, simulationId);
+        PreparedQuery percents = datastore.prepare(new Query(Schema.StatisticsState.entityKindBeaconsObservedPercent).setFilter(filterBySimulationId));
```
omermadmon, line 94: Prefer to split this line (one for the query and one for the prepared query).
(Same applies at `readDistancesStats`).

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
OrpazFishel: Otherwise we won't be able to read the simulation later - we should store this id somewhere.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
OrpazFishel: I don't want this PR to be too long, we can add it in another PR.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
omermadmon: Why not?
When writing the `SimulationMetadata` its key property is actually the simulationId. Whenever we want to retrieve a simulation Id it can be retrieved from the `SimulationMetadata` entity.

For example, when visualizing a simulation we first retrieve all `SimulationMetadata` entities from the db. Then the user chooses a simulation and we can retrieve the simulation's Id (which is the entity's key) and roundsNum in order to iteratively fetch the `ReadBoardState` servlet.

schainmariano (approved):
lgtm

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
OrpazFishel: I'll leave it for now and we'll decide later.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -68,15 +74,21 @@ void updateObserversAwaknessState() {
...
                 if (observer.isAwake()) {
                     double distance = distance(beacon.getLocation(), observer.getLocation());
                     if (distance <= transmissionThresholdRadius) {
                         observer.observe(transmission);
+                        observed = true; // for stats
                     }
                 }
             }
+            if (observed) {
+                beaconsObservedSum.put(String.valueOf(beacon.getId()),
```
nolaviz, line 89: Consider using `Map.merge` for this:

```
  beaconsObservedSum.merge(String.valueOf(beacon.getId()), 1.0D, Double::sum);
```

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -68,15 +74,21 @@ void updateObserversAwaknessState() {
      */
     void beaconsToObservers() {
         for (Beacon beacon : beacons) {
+            boolean observed = false; // for stats
```
nolaviz, line 77: Prefer to minimize comments; they are prone to rot (as it's easy to change code and forget to change the comment).
Instead, write the code in a way it will be clear even without the comment.
In this particular case, the comment just isn't needed - I can easily understand what `observed` *is*, and reading further down shows me it's used for stats just by its impact on them.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -94,10 +106,35 @@ void resolverEstimate() {
     void writeRoundState() { }
 
     /** Gather statistical data of the current round and update the aggregated simulation statistics based on all rounds. */
-    void updateSimulationStats() { }
+    void updateSimulationStats() {
+        Map<Beacon, Location> beaconsToEstimatedLocations = ((GlobalResolver) resolver).getBeaconsToEstimatedLocations();
+        List<Double> distances = beaconsToEstimatedLocations.keySet().stream() // ignore beacons that have never been observed
+                .map(beacon -> distance(beaconsToEstimatedLocations.get(beacon), beacon.getLocation())).collect(toList());
```
nolaviz, line 112: Prefer to use `toImmutableList()`, if possible. (Immutability is always preferred.)

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -94,10 +106,35 @@ void resolverEstimate() {
     void writeRoundState() { }
 
     /** Gather statistical data of the current round and update the aggregated simulation statistics based on all rounds. */
-    void updateSimulationStats() { }
+    void updateSimulationStats() {
+        Map<Beacon, Location> beaconsToEstimatedLocations = ((GlobalResolver) resolver).getBeaconsToEstimatedLocations();
```
nolaviz, line 110: Are you sure the cast here will always be valid? For all subclasses of the `AbstractSimulation`?
If so, consider just changing the field type.
Alternatively, add the method to the `IGlobalResolver` interface.
Avoid casts where you cannot definitely prove safety - otherwise they may manifest as runtime crashes. (In Google3 code, we have static analyzers that enforce this...)

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -94,10 +106,35 @@ void resolverEstimate() {
...
+        List<Double> distances = beaconsToEstimatedLocations.keySet().stream() // ignore beacons that have never been observed
+                .map(beacon -> distance(beaconsToEstimatedLocations.get(beacon), beacon.getLocation())).collect(toList());
+
+        if (!distances.isEmpty()) { // distances is empty until the first round an observer observed a beacon
+            double min = distances.stream().min(Double::compareTo).get();
+            distancesStats.put("min", Math.min(distancesStats.getOrDefault("min", Double.POSITIVE_INFINITY), min));
```
nolaviz, line 116: Consider using `Map.merge()` here (and in the following few rows) to simplify the code.

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -94,10 +106,35 @@ void resolverEstimate() {
...
+        }
+    }
 
     /** Write final simulation statistical data to db. */
-    void writeSimulationStats() { }
+    void writeSimulationStats() {
+        for (Beacon beacon : beacons) {
+            String beaconId = String.valueOf(beacon.getId());
+            beaconsObservedSum.putIfAbsent(beaconId, 0D);
+            beaconsObservedSum.put(beaconId, beaconsObservedSum.get(beaconId) / (currentRound - 1));
```
nolaviz, line 132: Why are you dividing by the round number?
(This actually does deserve a comment...)

src/main/java/com/google/research/bleth/simulator/StatisticsState.java
```diff
@@ -0,0 +1,125 @@
...
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.common.collect.ImmutableMap;
+import com.google.research.bleth.exceptions.StatisticsAlreadyExistException;
+import java.util.Map;
+
+/** A mediator between the statistics the simulation gathered and their storage on datastore. */
+public class StatisticsState {
+    private final String simulationId;
+    private final  Map<String, Double> distanceStats;
```
nolaviz, line 18: Nit: superfluous space before `Map`.

src/main/java/com/google/research/bleth/simulator/StatisticsState.java
```diff
@@ -0,0 +1,125 @@
...
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+        Map<String, Double> distanceStatsInDb = readDistancesStats(simulationId);
+        if (!distanceStatsInDb.isEmpty()) {
+            throw new StatisticsAlreadyExistException(simulationId);
+        }
+
+        Entity entity = new Entity(Schema.StatisticsState.entityKindDistance);
+        entity.setProperty(Schema.StatisticsState.simulationId, simulationId);
+        for (Map.Entry<String, Double> entry : distanceStats.entrySet()) {
```
nolaviz, line 49: Consider using `Map.forEach` here:

```
  distanceStats.forEach((k, v) -> entity.setProperty(k, v));
```

src/test/java/com/google/research/bleth/simulator/TracingSimulationIT.java
```diff
@@ -172,6 +176,212 @@ public void runTwoRoundsSimulationVerifyBeaconLocationOnEstimatedBoardAccordingT
...
+                .setAwakenessStrategyType(AwakenessStrategyFactory.Type.FIXED)
+                .build();
+
+        simulation.run();
+        String simulationId = simulation.getId();
+
+        Map<String, Double> observersStats = StatisticsState.readBeaconsObservedPercentStats(simulationId);
+
+        for (String beacon : observersStats.keySet()) {
+            assertThat(observersStats.get(beacon)).isEqualTo(0.5);
```
nolaviz, line 210: The "0.5" magic value here is surprising; please explain.

nolaviz (changes requested):

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -129,7 +128,8 @@ void writeSimulationStats() {
         for (Beacon beacon : beacons) {
             String beaconId = String.valueOf(beacon.getId());
             beaconsObservedSum.putIfAbsent(beaconId, 0D);
-            beaconsObservedSum.put(beaconId, beaconsObservedSum.get(beaconId) / (currentRound - 1));
+            beaconsObservedSum.put(beaconId, beaconsObservedSum.get(beaconId)
```
nolaviz, line 131: Consider splitting the computation to a separate variable (and then a properly-chosen variable name can obviate the need for a comment).

nolaviz (approved):

PR #25, opened 2020-11-22
Join Implementation

omermadmon:
A sort merge join implementation utilizing Datastore API. Will be used for retrieving all statistics entities corresponding to all simulations matching a certain condition (filter).

### DIFF
--- a/src/main/java/com/google/research/bleth/utils/Queries.java
+++ b/src/main/java/com/google/research/bleth/utils/Queries.java
@@ -16,7 +16,7 @@ import java.util.Optional;
 public class Queries {
 
     /**
-     * Given a primary entity kind and secondary entity kind, a foreign key and a filter, return a list of all entities
+     * Given a primary entity kind, a secondary entity kind, a foreign key and a filter, return a list of all entities
      * of the secondary entity matching an entity of the primary entity kind.
      *
      * Equivalent SQL syntax:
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/research/bleth/utils/Queries.java
```diff
@@ -0,0 +1,88 @@
...
+
+        // If one of these queries is empty, return an empty list.
+        if (!(primaryKeyIterator.hasNext() && secondaryEntityIterator.hasNext())) {
+            return result;
+        }
+
+        // Iterate over two queries results and add secondary entities with a matching foreign key.
+        String primaryKey = primaryKeyIterator.next();
+        Entity secondaryEntity = secondaryEntityIterator.next();
+        String secondaryKey = String.valueOf(secondaryEntity.getProperty(foreignKey));
```
nolaviz, line 63: Consider adding a helper method that does both of these steps together, so you don't need to duplicate them multiple times.

(Even better, introduce a second helper - with a similar name - that just wraps the `primaryKeyIterator.next()` call: then the code _here_ to advance both iterators would look the same.)

src/main/java/com/google/research/bleth/utils/Queries.java
```diff
@@ -0,0 +1,88 @@
...
+        while (primaryKeyIterator.hasNext() && secondaryEntityIterator.hasNext()) {
+            if (primaryKey.compareTo(secondaryKey) > 0) {
+                // primaryKey > secondaryKey
```
nolaviz, line 66: This comment isn't really needed: any sufficiently-experienced Java programmer knows the code in line 65 is the right way to write the condition in line 66.

nolaviz (approved):
Nice!

src/main/java/com/google/research/bleth/utils/Queries.java
omermadmon: I will leave it as is because of lines 74 and 83, as discussed.

PR #27, opened 2020-11-22
Speed Controller

omermadmon:


src/main/webapp/simulation_visualization.html
```diff
@@ -32,6 +32,15 @@ <h4>Estimated Board</h4>
             <table id="estimated-board-table" class="board-table"></table>
         </div>
     </div>
-    
+
+    <!-- Speed Controller. -->
+    <div class="speed-controller">
```
OrpazFishel, line 37: Looks great!
Nit: Please move the controller upward, above the boards - otherwise the user can miss it.

OrpazFishel (approved):

src/main/webapp/simulation_visualization.html
omermadmon: Good point, I'm on it :)

PR #28, opened 2020-11-23
Add to the cell's title the ids of all the beacons inside it; Color the cells that contain a specific beacon

OrpazFishel:
Screenshot 2020-11-24 at 11 57 59


omermadmon (commented):
Seems like a great solution for tracking after specific beacons.
Can you please provide a few screenshot? :)

omermadmon (approved):

PR #33, opened 2020-11-26
Add a description field when creating a new simulation

OrpazFishel:


src/main/webapp/new_simulation.html
```diff
@@ -22,6 +22,10 @@ <h1>Create and Run a New BLETH Simulation</h1>
     
     <!-- A form for a new simulation initialization. -->
     <form onload="setAllDynamicUpperBounds()" onchange="setAllDynamicUpperBounds()" id="newSimulationForm">
+        <!-- Description of the simulation. -->
+        <label for="simulationDescription">A short description of the simulation</label><br>
+        <input type="text" id="simulationDescription"><br>
+
```
omermadmon, line 28: Is this `description` might be relatively long? If so, we might want this to be a `textarea` tag (and if we do, its value should be manually added to the POST request parameters at `new_simulation.js`).

In any case, I think we should also set a `maxlength` attribute.

omermadmon (approved):
If the `description` might be relatively long we might not display it on the `simulations.html` table as a column. How about displaying only the `title` as a column, and only when hovering a simulation (row in the table) its description appears?

This can be done in a separate PR, of course.

PR #35, opened 2020-11-29
Delete Simulation Button

omermadmon:
Added a button for deleting a simulation from the db, by deleting all entities associated with its id (metadata, board states and stats).
image

### DIFF
--- a/src/main/java/com/google/research/bleth/servlets/DeleteSimulationServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/DeleteSimulationServlet.java
@@ -21,7 +21,7 @@ public class DeleteSimulationServlet extends HttpServlet {
         } catch (Exception e) {
             responseText = "Something went wrong: " + e.getMessage();
         }
-        response.setContentType("text/html;");
+        response.setContentType("text/plain;");
         response.getWriter().println(responseText);
     }
 }
--- a/src/main/java/com/google/research/bleth/servlets/DeleteSimulationServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/DeleteSimulationServlet.java
@@ -15,8 +15,13 @@ public class DeleteSimulationServlet extends HttpServlet {
     @Override
     public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
         String simulationId = request.getParameter("simulationId");
-        Queries.delete(simulationId);
+        String responseText = "Simulation has been deleted successfully.";
+        try {
+            Queries.delete(simulationId);
+        } catch (Exception e) {
+            responseText = "Something went wrong: " + e.getMessage();
+        }
         response.setContentType("text/html;");
-        response.getWriter().println("Simulation has been deleted successfully.");
+        response.getWriter().println(responseText);
     }
 }
--- a/src/main/java/com/google/research/bleth/servlets/DeleteSimulationServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/DeleteSimulationServlet.java
@@ -1,0 +1,22 @@
+package com.google.research.bleth.servlets;
+
+import com.google.research.bleth.utils.Queries;
+
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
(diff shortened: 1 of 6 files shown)


src/main/webapp/js/simulations.js
```diff
@@ -55,26 +55,53 @@ function addSimulationRows(table, simulations) {
...
-            // Redirect.
-            window.location.replace('simulation_visualization.html?' + toQueryString(simulationWithId));
-        });
-
-        row.insertCell(0).appendChild(visualizeSimulationButton);
+        const visualizeSimulationButton = createVisualizationButton(simulation, id);
+        const deleteSimulationButton = createDeletionButton(id);
+        const cell = row.insertCell(0);
+        cell.appendChild(visualizeSimulationButton);
+        cell.appendChild(deleteSimulationButton);
```
OrpazFishel, line 62: Aren't the buttons too close? We don't want the user to accidentally delete the simulation they want to see.

omermadmon: Clicking the delete button opens a confirmation box, so I think that's fine (this can be modified if visual design is the issue, of course).

src/main/java/com/google/research/bleth/servlets/DeleteSimulationServlet.java
```diff
@@ -0,0 +1,27 @@
...
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String simulationId = request.getParameter("simulationId");
+        String responseText = "Simulation has been deleted successfully.";
+        try {
+            Queries.delete(simulationId);
+        } catch (Exception e) {
+            responseText = "Something went wrong: " + e.getMessage();
+        }
+        response.setContentType("text/html;");
+        response.getWriter().println(responseText);
```
nolaviz, line 25: This is generating invalid HTML - (1) you're missing (at minimum) <html><body> ... </body></html>, (2) if the exception message includes any special characters they won't be escaped (and can confuse the parsers even further).

Either change the response type to text/plain, or properly wrap and HTML-escape your output.

src/main/java/com/google/research/bleth/utils/Queries.java
```diff
@@ -116,4 +117,25 @@ public static double average(List<Entity> entities, String property) throws Clas
         return properties.stream()
                 .collect(toImmutableMap(Function.identity(), property -> average(entities, property)));
     }
+
+    /**
+     * Given a simulationId as a string, delete all data associated with that simulation id.
+     * @param simulationId is the simulation id.
+     */
+    public static void delete(String simulationId) {
```
nolaviz, line 125: Does DataStore support transactions? If so, can you wrap all these deletions in a single transaction (so a partial failure won't leave our database in a weird state)?

src/main/java/com/google/research/bleth/utils/Queries.java
```diff
@@ -116,4 +117,25 @@ public static double average(List<Entity> entities, String property) throws Clas
...
+        delete(Schema.BoardState.entityKindReal, Schema.BoardState.simulationId, simulationId);
+        delete(Schema.BoardState.entityKindEstimated, Schema.BoardState.simulationId, simulationId);
+        datastore.delete(KeyFactory.stringToKey(simulationId));
+    }
+
+    private static void delete(String entityKind, String foreignKeyProperty, String foreignKeyKeyValue) {
+        Query.Filter filter = new Query.FilterPredicate(foreignKeyProperty, Query.FilterOperator.EQUAL, foreignKeyKeyValue);
+        Query entitiesToDelete = new Query(entityKind).setFilter(filter);
+        Iterable<Entity> entitiesToDeleteIterable = datastore.prepare(entitiesToDelete).asIterable();
+        for (Entity entity : entitiesToDeleteIterable) {
```
nolaviz, line 137: This loop is not strictly needed: `DataStoreService` has a `delete` method accepting an `Iterable` directly. (You can transform the `Iterable<Entity>` to an `Iterable<Key>` using Guava `Iterables.transform()`)

nolaviz (approved):

PR #38, opened 2020-11-30
Add a pause button

OrpazFishel:


src/main/webapp/js/simulation_visualization.js
```diff
@@ -124,4 +129,14 @@ function extractBeaconsIds(agents) {
...
+}
+
+/** Change the button's content from Pause to Play or vice versa. */
+function switchState() {
+    var pauseButton = document.getElementById('pause-button');
+    if (pauseButton.innerHTML === "Pause") {
+        pauseButton.innerHTML = "Play";
+    } else {
+        pauseButton.innerHTML = "Pause"
+    }
```
omermadmon, line 141: Try conditional assignment:
```js
pauseButton.innerHTML =  pauseButton.innerHTML === "Pause" ? "Play" : "Pause";
```

src/main/webapp/js/simulation_visualization.js
```diff
@@ -1,6 +1,7 @@
 import { toQueryString, getUrlVars, sleep } from './utils.js';
 
 window.visualize = visualize; // Add function to global scope.
+window.switchState = switchState; // Add function to global scope.
```
omermadmon, line 4: Consider replacing those two comments with a single comment before these two lines.

src/main/webapp/js/simulation_visualization.js
```diff
@@ -13,10 +15,14 @@ async function visualize() {
     var mainHeader = document.getElementById('simulation-visualization-header');
     var roundHeader = document.getElementById('current-round-header');
     var delay;
+    var pauseButton = document.getElementById('pause-button');
     mainHeader.innerText = 'Visualizing Simulation\n' + simulation.id;
     var params = {};
     params['simulationId'] = simulation.id;
     for (var round = 0; round < simulation.roundsNum; round++) {
+        while (pauseButton.innerHTML === "Play") {
```
nolaviz, line 23: Consider adding a function ("isPaused") to wrap this condition - otherwise I find it hard to guess what the condition means.

nolaviz (approved):
Looks good to me, except one small comment.
What was the issue requiring my attention?

omermadmon:
> Looks good to me, except one small comment.
> What was the issue requiring my attention?

I actually wasn't sure about lines 22-24:
```js
while (isPaused()) {
            await sleep(10);
        }
```
Is there a more elegant way to pause than 'sleep while true'?

nolaviz:
Tzvika - can you please help / advise?
My interns need a "pause" button in their (HTML+JS) app; how would you
recommend they implement this?

nolaviz:
Can you send me more context? How do they run when unpaused?
Do you use setInterval to proceed between steps? If you do you can cancel
the interval on pause and restart it on start.
If there is some other mechanism then I need to see the code. But busy wait
is likely not a good solution.

nolaviz:
Looking at the code, I think the for loop body should be moved to a
function, let's call it *nextRound*.
When the function ends instead of calling sleep it can use setTimeout to
run the next iteration.

// This is the timeout variable.
var simulationTimeout;

// Holds the simulation state including loop count and other parameters.
var state = {
    "paused": false,
    "finished": false, // set to true on last iteration
};

// Starts/resumes the simulation.
function start() {
    state.paused = false;
    simulationTimeout = setTimeout(0, nextRound);
}

// Pauses the simulation by not running the next timeout.
// Keep in mind JS is single threaded so there is no race condition I
believe between this and nextRound setting a new timeout.
function pause() {
    state.paused = true;
    clearTimeout(simulationTimeout);
}

function nextRound() {
    /// do logic.
    if (!state.paused && !state.finished) {
        simulationTimeout = setTimeout(delay, nextRound);
    }
}

PR #39, opened 2020-12-02
Tasks Queue

omermadmon:
Implemented a servlet to receive POST request from client and enqueue a task into `simulations-queue`.
This lead to the following performance improvements:
* A 25*25 simulations runs for about 500 rounds.
* A 50*50 simulations runs for about 150 rounds. 

To be done next:
* Configure tasks queue timeout (currently 10 mins).
* Wrap simulation build and run in a single transaction.
* Limit number of rounds in the new simulation form.

### DIFF
--- a/src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
@@ -46,7 +46,7 @@ public class EnqueueSimulationServlet extends HttpServlet {
             client.createTask(queueName, task);
         }
 
-        response.setContentType("text/html;");
+        response.setContentType("text/plain;");
         response.getWriter().println("Task has been added to queue.");
     }
 
--- a/src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
@@ -14,6 +14,10 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+/**
+ * A servlet used for enqueuing a task targeted at endpoint '/new-simulation',
+ * in order to create and run a new simulation.
+ */
 @WebServlet("/enqueue-simulation")
 public class EnqueueSimulationServlet extends HttpServlet {
 
--- a/src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
@@ -33,28 +33,29 @@ public class EnqueueSimulationServlet extends HttpServlet {
             RequestDispatcher dispatcher = getServletContext()
                     .getRequestDispatcher("/new-simulation");
             dispatcher.forward(request, response);
-        } else {
-            try (CloudTasksClient client = CloudTasksClient.create()) {
-                // Construct the HTTP request (for the task).
-                AppEngineHttpRequest httpRequest = AppEngineHttpRequest.newBuilder()
-                        .setRelativeUri("/new-simulation")
-                        .setHttpMethod(HttpMethod.POST)
-                        .putHeaders("Content-Type", request.getContentType())
-                        .setBody(ByteString.readFrom(request.getInputStream()))
-                        .build();
-
-                // Construct the task body.
-                Task task = Task.newBuilder()
-                        .setAppEngineHttpRequest(httpRequest)
-                        .build();
-
-                // Add the task to the queue.
-                String queueName = QueueName.of(PROJECT_ID, LOCATION_ID, QUEUE_ID).toString();
-                client.createTask(queueName, task);
-            }
-
-            response.setContentType("text/plain;");
-            response.getWriter().println("Task has been added to queue.");
+            return;
         }
+
+        try (CloudTasksClient client = CloudTasksClient.create()) {
+            // Construct the HTTP request (for the task).
+            AppEngineHttpRequest httpRequest = AppEngineHttpRequest.newBuilder()
+                    .setRelativeUri("/new-simulation")
+                    .setHttpMethod(HttpMethod.POST)
+                    .putHeaders("Content-Type", request.getContentType())
+                    .setBody(ByteString.readFrom(request.getInputStream()))
+                    .build();
+
+            // Construct the task body.
+            Task task = Task.newBuilder()
+                    .setAppEngineHttpRequest(httpRequest)
+                    .build();
+
+            // Add the task to the queue.
+            String queueName = QueueName.of(PROJECT_ID, LOCATION_ID, QUEUE_ID).toString();
+            client.createTask(queueName, task);
+        }
+
+        response.setContentType("text/plain;");
+        response.getWriter().println("Task has been added to queue.");
     }
 }
--- a/src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
@@ -7,6 +7,8 @@ import com.google.cloud.tasks.v2.QueueName;
 import com.google.cloud.tasks.v2.Task;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/enqueue-simulation")
```
OrpazFishel, line 17: Please add Javadoc.

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
```diff
@@ -0,0 +1,60 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * A servlet used for enqueuing a task targeted at endpoint '/new-simulation',
+ * in order to create and run a new simulation.
+ */
+@WebServlet("/enqueue-simulation")
+public class EnqueueSimulationServlet extends HttpServlet {
+
+    static final String projectId = "bleth-2020";
```
nolaviz, line 24: You can get this from `com.google.appengine.api.utils.SystemProperty.applicationId()`, AFAIK.

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
```diff
@@ -0,0 +1,60 @@
...
+    static final String locationId = "europe-west1";
+    static final String queueId = "simulations-queue";
+
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+        try (CloudTasksClient client = CloudTasksClient.create()) {
+            // Construct the fully qualified queue name.
+            String queueName = QueueName.of(projectId, locationId, queueId).toString();
```
nolaviz, line 32: Can this be changed into a class-wide constant? (It seems to only depend on other constants?)

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
nolaviz, line 24: Please use UPPER_SNAKE_CASE for constant names.

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
```diff
@@ -0,0 +1,60 @@
...
+            // Construct the task body.
+            Task task = Task.newBuilder()
+                    .setAppEngineHttpRequest(httpRequest)
+                    .build();
+
+            // Add the task to the queue.
+            client.createTask(queueName, task);
+        }
+
+        response.setContentType("text/html;");
```
nolaviz, line 49: text/plain?
Or change the content to actually be HTML...?

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
```diff
@@ -0,0 +1,60 @@
...
+        response.getWriter().println("Task has been added to queue.");
+    }
+
+    private String toQueryString(HttpServletRequest request) {
```
nolaviz, line 53: Why do you need this? Why not just use `request.getQueryString()`?
(AFAIK `request.getParameterMap()` just parses the query string.)

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
```diff
@@ -0,0 +1,60 @@
...
+        List<String> keyValuePairs = new ArrayList<>();
+        for (String param : request.getParameterMap().keySet()) {
+            keyValuePairs.add(param + "=" + request.getParameter(param));
```
nolaviz, line 56: Parameter values in a query string need to be URL-escaped.

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
omermadmon: This POST requests carries its parameters within the request body, which is why `request.getQueryString()` returns `null` in that case.

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
omermadmon: I thought of using [URLEncoder](https://docs.oracle.com/javase/8/docs/api/java/net/URLEncoder.html#encode-java.lang.String-java.lang.String-):
```java
keyValuePairs.add(URLEncoder.encode(param) + "=" + URLEncoder.encode(request.getParameter(param)));
```
Is that what you mean?

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
nolaviz: In this case - why move the parameters from the body to the URL at all?

IMO it would be preferable to copy the POST body as-is.

```
AppEngineHttpRequest httpRequest = AppEngineHttpRequest.newBuilder()
                .setRelativeUri("/new-simulation?" + toQueryString(request))
                .setHttpMethod(HttpMethod.POST)
                .putHeaders("Content-Type", request.getContentType())
                .setPayload(getRequestPayload(request))
                .build();

// ...

  }

// ...

  private static byte[] getRequestPayload(HttpServletRequest request) throws IOException {
    try (InputStream s = request.getInputStream()) {
      return s.readAllBytes();
    }
  }
```

omermadmon: I have three questions:
- If we pass the request parameters within its body, what do we need `putHeaders()` for? 
- Is there a difference between `setBody()` and `setPayload()`?
- If we do so, we don't have to add the query string at `setRelativeUri()`, right?

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
omermadmon: The following exception is thrown: `com.google.api.gax.rpc.UnavailableException: io.grpc.StatusRuntimeException: UNAVAILABLE: Credentials failed to obtain metadata`
I have tried to figure it out and it seems like a possible cause might be incorrect timezones in containerized / virtualized environments (I might be wrong though). 
Isn't it safer to just hard code it, or define it somewhere inside the project itself?

omermadmon:
@nolaviz 
I have added a condition at lines 32-35 to check if the servlet runs on localhost, and if so redirect the request without using the queue. It doesn't feel like the best way to implement this logic so I would love to get a feedback on that as well. Thanks.

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
nolaviz: I suspect you can only call this from within a running servlet, not from initialization code (i.e. it cannot be a constant).

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
nolaviz: * We need `putHeaders("Content-Type", ...)` since the body can only be interpreted in the context of a specific content-type.
* `setBody` is right, and `setPayload` is me being confused.
* Right - no need for a query string.

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
nolaviz: Using `URLEncoder` is indeed the right way to do it.

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
```diff
@@ -0,0 +1,60 @@
...
+    static final String QUEUE_ID = "simulations-queue";
+
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException {
+
+        // If servlet runs on localhost, forward request without creating and enqueuing a task.
+        if (request.getServerName().equals("localhost") && request.getServerPort() == 8080) {
+            RequestDispatcher dispatcher = getServletContext()
+                    .getRequestDispatcher("/new-simulation");
+            dispatcher.forward(request, response);
+        } else {
```
nolaviz, line 36: Prefer to just add a "return" into the "then" block, so the "else" part can just be unconditional.

This makes the code easier to read - a reader can know that the localhost case is completely done, and there's nothing later (outside the conditional) that still needs to happen. So they can now concentrate fully on the regular path.

src/main/java/com/google/research/bleth/servlets/EnqueueSimulationServlet.java
nolaviz: Yes, using `UrlEncoder` this way is the right way to do it.

nolaviz (approved):

PR #40, opened 2020-12-06
 Change the visualization to be based on timeout instead of for loop, add a pause button

OrpazFishel:


src/main/webapp/js/simulation_visualization.js
```diff
@@ -1,38 +1,97 @@
 import { toQueryString, getUrlVars, sleep } from './utils.js';
 
-window.visualize = visualize; // Add function to global scope.
+window.visualize = startVisualization; // Add function to global scope.
```
omermadmon, line 3: Please move this comment below the code line (to keep consistency).

omermadmon (approved):

PR #46, opened 2020-12-08
Sort Simulations Table

omermadmon:
Provided buttons for sorting the simulations table by each column.
First click -> ascending sort, second click -> descending order.
image

### DIFF
--- a/src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
@@ -15,6 +15,7 @@
 package com.google.research.bleth.servlets;
 
 import com.google.appengine.api.datastore.Query;
+import com.google.appengine.repackaged.com.google.common.base.Ascii;
 import com.google.gson.Gson;
 import com.google.gson.JsonElement;
 import com.google.research.bleth.simulator.SimulationMetadata;
@@ -41,8 +42,8 @@ public class ListSimulationsServlet extends HttpServlet {
             sortProperty = Optional.of(request.getParameter("sortProperty"));
         }
         if (request.getParameter("sortDirection") != null) {
-            int sortDirectionIndex = Integer.parseInt(request.getParameter("sortDirection"));
-            sortDirection = Optional.ofNullable(Query.SortDirection.values()[sortDirectionIndex]);
+            String sortDirectionString = Ascii.toUpperCase(request.getParameter("sortDirection"));
+            sortDirection = Optional.of(Query.SortDirection.valueOf(sortDirectionString));
         }
 
         LinkedHashMap<String, JsonElement> simulationsAsJson = new LinkedHashMap<>();
--- a/src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
@@ -16,13 +16,13 @@ package com.google.research.bleth.servlets;
 
 import com.google.appengine.api.datastore.Query;
 import com.google.appengine.repackaged.com.google.common.base.Ascii;
+import com.google.common.collect.ImmutableMap;
 import com.google.gson.Gson;
 import com.google.gson.JsonElement;
 import com.google.research.bleth.simulator.SimulationMetadata;
 import com.google.research.bleth.utils.Queries;
 
 import java.io.IOException;
-import java.util.LinkedHashMap;
 import java.util.Optional;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
@@ -45,11 +45,11 @@ public class ListSimulationsServlet extends HttpServlet {
             sortingParameters = Optional.of(new Queries.SortingParameters(sortProperty, sortDirection));
         }
 
-        LinkedHashMap<String, JsonElement> simulationsAsJson = new LinkedHashMap<>();
+        ImmutableMap.Builder<String, JsonElement> simulationsAsJson = new ImmutableMap.Builder<>();
         SimulationMetadata.listSimulations(sortingParameters)
                 .forEach((id, metadata) -> simulationsAsJson.put(id, gson.toJsonTree(metadata)));
 
         response.setContentType("application/json;");
-        response.getWriter().println(gson.toJson(simulationsAsJson));
+        response.getWriter().println(gson.toJson(simulationsAsJson.build()));
     }
 }
--- a/src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
@@ -14,6 +14,7 @@
 
 package com.google.research.bleth.servlets;
 
+import com.google.appengine.api.datastore.Query;
 import com.google.gson.Gson;
 import com.google.gson.JsonElement;
 import com.google.research.bleth.simulator.SimulationMetadata;
@@ -36,10 +37,15 @@ public class ListSimulationsServlet extends HttpServlet {
 
         // Read SimulationMetadata as a HashMap.
         Optional<String> sortProperty = Optional.empty();
+        Optional<Query.SortDirection> sortDirection = Optional.empty();
         if (request.getParameter("sortProperty") != null) {
             sortProperty = Optional.of(request.getParameter("sortProperty"));
         }
-        LinkedHashMap<String, SimulationMetadata> simulations = SimulationMetadata.listSimulations(sortProperty);
+        if (request.getParameter("sortDirection") != null) {
+            int sortDirectionIndex = Integer.parseInt(request.getParameter("sortDirection"));
+            sortDirection = Optional.ofNullable(Query.SortDirection.values()[sortDirectionIndex]);
+        }
+        LinkedHashMap<String, SimulationMetadata> simulations = SimulationMetadata.listSimulations(sortProperty, sortDirection);
 
         // Serialize SimulationMetadata objects to JSON strings.
         LinkedHashMap<String, JsonElement> simulationsAsJson = new LinkedHashMap<>();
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
```diff
@@ -35,7 +35,7 @@
     public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
         Gson gson = new Gson(); // Used for json serialization.
 
-        // Read SimulationMetadata as a HashMap.
+        // Read SimulationMetadata as a sorted LinkedHashMap.
```
schainmariano, line 38: why comment at all.. ?

nolaviz: +1 : Google best practices say

> Use a comment when it is infeasible to make your code self-explanatory

src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
```diff
@@ -34,11 +35,20 @@
...
-        // Read SimulationMetadata as a HashMap.
-        ImmutableMap<String, SimulationMetadata> simulations = SimulationMetadata.listSimulations();
+        // Read SimulationMetadata as a sorted LinkedHashMap.
+        Optional<String> sortProperty = Optional.empty();
+        Optional<Query.SortDirection> sortDirection = Optional.empty();
+        if (request.getParameter("sortProperty") != null) {
+            sortProperty = Optional.of(request.getParameter("sortProperty"));
+        }
+        if (request.getParameter("sortDirection") != null) {
+            int sortDirectionIndex = Integer.parseInt(request.getParameter("sortDirection"));
```
nolaviz, line 45: Why are you passing this as an integer?
Prefer to pass the direction as a string ("ASCENDING" or "DESCENDING"), and use `SortDirection.valueOf(Ascii.toUpperCase(request.getParameter("sortDirection")))`. This way the request URL is easier to understand by us humans (e.g. in logs, when debugging, etc.).

src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
```diff
@@ -34,11 +35,20 @@
...
 
-        // Read SimulationMetadata as a HashMap.
-        ImmutableMap<String, SimulationMetadata> simulations = SimulationMetadata.listSimulations();
+        // Read SimulationMetadata as a sorted LinkedHashMap.
+        Optional<String> sortProperty = Optional.empty();
+        Optional<Query.SortDirection> sortDirection = Optional.empty();
+        if (request.getParameter("sortProperty") != null) {
+            sortProperty = Optional.of(request.getParameter("sortProperty"));
+        }
+        if (request.getParameter("sortDirection") != null) {
```
nolaviz, line 44: Having sortDirection set without a sortProperty is meaningless (and is likely a violation of this servlet's API).
Prefer to return an error if this is the case.

src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
```diff
@@ -34,11 +35,20 @@
     public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
         Gson gson = new Gson(); // Used for json serialization.
 
-        // Read SimulationMetadata as a HashMap.
-        ImmutableMap<String, SimulationMetadata> simulations = SimulationMetadata.listSimulations();
+        // Read SimulationMetadata as a sorted LinkedHashMap.
+        Optional<String> sortProperty = Optional.empty();
```
nolaviz, line 39: Since sortProperty and sortDirection are always used together, prefer to store them both in a single helper object, e.g.

```
static class SortingParameters {
  private String sortBy;
  private Query.SortDirection direction = Query.SortDirection.ASCENDING;

  String sortBy() { return sortBy; }
  Query.SortDirection direction() { return direction; }
}
```

(this way you also only pass a single optional parameter to `listSimulations()`, *and* you don't need to have separate special cases when `sortDirection` is specified explicitly or not.)

src/main/java/com/google/research/bleth/simulator/SimulationMetadata.java
```diff
@@ -108,17 +110,27 @@ public static SimulationMetadata read(String simulationId) {
 
     /**
      * Read all existing SimulationMetadata entites from the db, and return them as a hashmap.
+     * @param sortProperty is the name of the property to sort the results by (optional).
+     * @param sortDirection is the direction of the sort (optional).
      * @return a hashmap mapping a simulationId to the corresponding SimulationMetadata object.
      */
-    public static ImmutableMap<String, SimulationMetadata> listSimulations() {
-        ImmutableMap.Builder<String, SimulationMetadata> simulations = new ImmutableMap.Builder<>();
+    public static LinkedHashMap<String, SimulationMetadata> listSimulations(Optional<String> sortProperty,
```
nolaviz, line 117: Why is this a `Map` at all? Does any caller ever invoke `.get(foo)` on the returned value? (The only caller I see is the servlet, and it only iterates over the results)

Prefer to just change this method to return a `List<SimulationMetadata>`, which is precisely what the caller actually uses this for.

src/main/java/com/google/research/bleth/simulator/SimulationMetadata.java
```diff
@@ -108,17 +110,27 @@ public static SimulationMetadata read(String simulationId) {
...
+                                                                            Optional<Query.SortDirection> sortDirection) {
+        LinkedHashMap<String, SimulationMetadata> simulations = new LinkedHashMap<>();
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         Query simulationMetadataQuery = new Query(Schema.SimulationMetadata.entityKind);
+        if (sortProperty.isPresent()) {
+            if (sortDirection.isPresent()) {
```
nolaviz, line 123: As mentioned in the caller - prefer to just have the caller default `sortDirection` to `ASCENDING`, and then you don't need this extra "if" at all.

src/main/webapp/js/simulations.js
```diff
@@ -15,6 +15,10 @@
 import { toQueryString } from './utils.js';
 
 window.retrieveSimulations = retrieveSimulations; // Add function to global scope.
+const ASC = 0; // Query.SortDirection.ASCENDING = 0 (according to Datastore API documentation).
```
nolaviz, line 18: Relying on the ordering of enums is rather brittle; as mentioned in my comment on the servlet, prefer to just pass "ASCENDING" or "DESCENDING" as strings in the URL.

nolaviz (changes requested):

src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
omermadmon: Do we have to throw an exception in such case if we are using the `SortingParameters` you have suggested? I thought of constructing a `SortingParameters` optional only if both parameters are not null, and otherwise return the results unsorted (in practice, either both are provided or neither is provided).

src/main/java/com/google/research/bleth/simulator/SimulationMetadata.java
omermadmon: The reason is that the id is not a property of the `SimulationMetadata` class; the id is the return value of the `SimulationMetadata.write()` method. In order to apply the change you have suggested, we need to add an `id` property to the `SimulationMetadata` class, and update it inside the `SimulationMetadata.write()`.

This change sound relatively simple, but I wonder if the id _should_ actually be a property of the `SimulationMetadata` class - what should happen if the same `SimulationMetadata` object calls his `write()` method multiple time? This might happen if we use the same builder to construct multiple simulation with the same setup, for example.

omermadmon: Plus: this key-value struct is being used in client-side code, where this map is being parsed as a JSON object (a list can also be parsed as JSON, but it is easier to deal with when the key is the simulation id and the value is another JSON object storing all other simulation's metadata parameters).

src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
nolaviz: It's not a must, but - it is a better API. (In theory, if someone else were to write a new frontend for your service, they could be surprised at this behavior).
Your call.

omermadmon: So this means we're treating differently each one of the four possible scenarios: (property, direction) X (provided, not provided), right?

src/main/java/com/google/research/bleth/simulator/SimulationMetadata.java
nolaviz: Still - at least here, you're not truly using this as a "map" per se; it's more of a "list of key-value pairs".

My usual recommendation in this case would be to create yet another class that holds a String key and a `SimulationMetadata` element, and return a list of those.

omermadmon: No problem at all, but may I ask why? Isn't this exactly what `LinkedHashMap` does?

nolaviz: `LinkedHashMap` is still a `Map`, and the assumption of the reader is that this is important (i.e. that callers need to use the `Map`-specific APIs, such as `get(key)` and `put(key, value)`. Here you're just using this class for its convenience value of having an `entrySet()` - which isn't a core part of what a `Map` is.

src/main/java/com/google/research/bleth/servlets/ListSimulationsServlet.java
nolaviz: Possibly.

Or just - create the `SortingParameters` if both parameters are set, return an error if just one is set, return unsorted result if no parameters are set.

src/main/java/com/google/research/bleth/simulator/SimulationMetadata.java
nolaviz: OK; I have to eat a hat.

I just looked at the internal guide to "choosing simpler types" - and it does recommend to use a `Map` instead of an `Iterable<Pair<Foo, Bar>>` or equivalent.

One thing it does mention is that `ImmutableMap` *also* provides deterministic ordering (i.e., its `entrySet` is ordered in the same order of elements added to the `builder()`). So - just use `ImmutableMap`, and be done with it.

¯\_(ツ)_/¯

nolaviz (approved):

PR #53, opened 2021-01-03
Experiment Feature

omermadmon:
- `EnqueueExperimentServlet` receives an HTTP request from client, create an `Experiment` Datastore entity, generates simulation parameters configurations and enqueues multiple HTTP tasks.
- `NewExperimentSimulationServlet` dequeues a single tasks from the Tasks Queue, creates and runs a new simulation, appends an `ExperimentsToSimulations` Datastore entity and updates its associated `Experiment` Datastore entity.

### DIFF
--- a/src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
@@ -1,0 +1,104 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.research.bleth.servlets;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+import com.google.gson.Gson;
+
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * A servlet used for enqueuing multiple tasks targeted at endpoint '/new-experiment-simulation',
+ * in order to create and run a new experiment.
+ */
+@WebServlet("/enqueue-experiment")
+public class EnqueueExperimentServlet extends HttpServlet {
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+        String experimentTitle = request.getParameter("experimentTitle");
+        int beaconsNum = Integer.parseInt(request.getParameter("beaconsNum"));
+        Set<PropertyWrapper> roundsNumValues = createIntValues(request, "roundsNum");
+        Set<PropertyWrapper> rowsNumValues = createIntValues(request, "rowsNum");
+        Set<PropertyWrapper> colsNumValues = createIntValues(request, "colsNum");
+        Set<PropertyWrapper> observersNumValues = createIntValues(request, "observersNum");
+        Set<PropertyWrapper> awakenessCycleValues = createIntValues(request, "awakenessCycle");
+        Set<PropertyWrapper> awakenessDurationValues = createIntValues(request, "awakenessDuration");
+        Set<PropertyWrapper> transmissionThresholdRadiusValues = createDoubleValues(request, "transmissionThresholdRadius");
+
+        Set<List<PropertyWrapper>> configurations = Sets.cartesianProduct(
+                roundsNumValues,
+                rowsNumValues,
+                colsNumValues,
+                observersNumValues,
+                awakenessCycleValues,
+                awakenessDurationValues,
+                transmissionThresholdRadiusValues
+        );
+
+        // todo: for p in product:
+            // todo: extract p's parameters to create an http request
+            // todo: create a task and enqueue it
+
+        String testOutput = new Gson().toJson(configurations);
+        response.setContentType("application/json;");
+        response.getWriter().println(testOutput);
+    }
+
+    @AutoValue
+    public static abstract class PropertyWrapper {
+        public static PropertyWrapper create(String property, Number value) {
+            return new AutoValue_EnqueueExperimentServlet_PropertyWrapper(property, value);
+        }
+        public abstract String property();
+        public abstract Number value();
+    }
+
+    private Set<PropertyWrapper> createIntValues(HttpServletRequest request, String parameter) {
+        int lower = Integer.parseInt(request.getParameter("lower" + upperCaseFirstChar(parameter)));
+        int upper = Integer.parseInt(request.getParameter("upper" + upperCaseFirstChar(parameter)));
+        int step = Integer.parseInt(request.getParameter("step" + upperCaseFirstChar(parameter)));
+        Set<PropertyWrapper> values = new HashSet<>();
+        for (int value = lower; value <= upper; value += step) {
+            values.add(PropertyWrapper.create(parameter, value));
+        }
+        return ImmutableSet.copyOf(values);
+    }
+
+    private Set<PropertyWrapper> createDoubleValues(HttpServletRequest request, String parameter) {
+        double lower = Double.parseDouble(request.getParameter("lower" + upperCaseFirstChar(parameter)));
+        double upper = Double.parseDouble(request.getParameter("upper" + upperCaseFirstChar(parameter)));
+        double step = Double.parseDouble(request.getParameter("step" + upperCaseFirstChar(parameter)));
+        Set<PropertyWrapper> values = new HashSet<>();
+        for (double value = lower; value <= upper; value += step) {
+            values.add(PropertyWrapper.create(parameter, value));
+        }
+        return ImmutableSet.copyOf(values);
+    }
+
+    private String upperCaseFirstChar(String s) {
+        return s.substring(0, 1).toUpperCase() + s.substring(1);
+    }
+}
--- a/src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
@@ -86,7 +86,7 @@ public class EnqueueExperimentServlet extends HttpServlet {
         }
 
         response.setContentType("text/plain;");
-        response.getWriter().println("All tasks have been added to queue.");
+        response.getWriter().println(legalConfigurationsCount + " tasks have been added to queue.");
     }
 
     @AutoValue
--- a/src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
@@ -209,7 +209,7 @@ public class EnqueueExperimentServlet extends HttpServlet {
                 .collect(Collectors.toMap(PropertyWrapper::property, PropertyWrapper::value));
 
         boolean legalDimensions = properties.get(Schema.SimulationMetadata.rowsNum).intValue() > 0 &&
-                properties.get(Schema.SimulationMetadata.rowsNum).intValue() > 0;
+                properties.get(Schema.SimulationMetadata.colsNum).intValue() > 0;
         boolean legalAgentsNumber = properties.get(Schema.SimulationMetadata.observersNum).intValue() > 0;
         boolean legalCycleAndDuration = properties.get(Schema.SimulationMetadata.awakenessCycle).intValue() > 0 &&
                 properties.get(Schema.SimulationMetadata.awakenessDuration).intValue() > 0;
--- a/src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
@@ -14,6 +14,8 @@
 
 package com.google.research.bleth.servlets;
 
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
@@ -130,7 +132,7 @@ public class EnqueueExperimentServlet extends HttpServlet {
                 transmissionThresholdRadiusValues
         );
 
-        return configurations.stream().filter(this::validateArguments).collect(ImmutableSet.toImmutableSet());
+        return configurations.stream().filter(this::validateArguments).collect(toImmutableSet());
     }
 
     private Set<PropertyWrapper> createIntValues(HttpServletRequest request, String parameter) {
--- a/src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
+++ b/src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
@@ -145,9 +145,9 @@ public class EnqueueExperimentServlet extends HttpServlet {
     }
 
     private Set<PropertyWrapper> createIntValues(HttpServletRequest request, String parameter) {
-        int lower = Integer.parseInt(request.getParameter("lower" + upperCaseFirstChar(parameter)));
-        int upper = Integer.parseInt(request.getParameter("upper" + upperCaseFirstChar(parameter)));
-        int step = Integer.parseInt(request.getParameter("step" + upperCaseFirstChar(parameter)));
+        int lower = Integer.parseInt(request.getParameter("lower" + capitalizeFirstChar(parameter)));
+        int upper = Integer.parseInt(request.getParameter("upper" + capitalizeFirstChar(parameter)));
+        int step = Integer.parseInt(request.getParameter("step" + capitalizeFirstChar(parameter)));
         Set<PropertyWrapper> values = new HashSet<>();
         for (int value = lower; value <= upper; value += step) {
             values.add(PropertyWrapper.create(parameter, value));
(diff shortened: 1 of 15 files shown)


src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
```diff
@@ -0,0 +1,233 @@
...
+
+        Entity experiment = new Entity(Schema.Experiment.entityKind);
+        experiment.setProperty(Schema.Experiment.experimentTitle, experimentTitle);
+        Key experimentId = datastore.put(experiment);
+        log.info("A new experiment entity with id " + KeyFactory.keyToString(experimentId) +
+                "was created and written to db.");
+
+        int legalConfigurationsCount = 0;
+        for (List<PropertyWrapper> configuration : configurations) {
+            if (validateArguments(configuration)) {
```
OrpazFishel, line 94: Consider checking this out before, i.e. filtering the Cartesian product so that the assistant function returns only valid configurations. If you do so you don't have to count the new tasks here.

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
```diff
@@ -0,0 +1,233 @@
...
+            response.setContentType("text/plain;");
+            response.getWriter().println(e.getMessage());
+        }
+
+        response.setContentType("text/plain;");
+        response.getWriter().println(legalConfigurationsCount + " tasks have been added to queue.");
+    }
+
+    @AutoValue
+    public static abstract class PropertyWrapper {
```
OrpazFishel, line 121: Please add Javadoc.

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
```diff
@@ -0,0 +1,233 @@
...
+                rowsNumValues,
+                colsNumValues,
+                observersNumValues,
+                awakenessCycleValues,
+                awakenessDurationValues,
+                transmissionThresholdRadiusValues
+        );
+    }
+
+    private Set<PropertyWrapper> createIntValues(HttpServletRequest request, String parameter) {
```
OrpazFishel, line 149: Is this function necessary? It seems like createDoubleValues is sufficient.

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
```diff
@@ -0,0 +1,233 @@
...
+    private String toQueryString(Map<String, String> params) throws UnsupportedEncodingException {
+        List<String> keyValuePairs = new ArrayList<>();
+        for (String param : params.keySet()) {
+            keyValuePairs.add(URLEncoder.encode(param, StandardCharsets.UTF_8.toString()) + "=" +
+                    URLEncoder.encode(params.get(param), StandardCharsets.UTF_8.toString()));
+        }
+        return String.join("&", keyValuePairs);
+    }
+
+    private boolean validateArguments(List<PropertyWrapper> configuration) {
```
OrpazFishel, line 206: Consider phrase this via the positive conditions, rather than the negative; i.e. check if the args are legal.

src/main/java/com/google/research/bleth/servlets/NewExperimentSimulationServlet.java
```diff
@@ -0,0 +1,138 @@
...
+                    .setObserverMovementStrategyType(observerMovementStrategy)
+                    .setAwakenessStrategyType(observerAwakenessStrategy)
+                    .setAwakenessCycle(awakenessCycle)
+                    .setAwakenessDuration(awakenessDuration)
+                    .setTransmissionThresholdRadius(transmissionThresholdRadius)
+                    .build();
+
+            // Run simulation.
+            simulation.run();
+            simulationId = simulation.getId();
```
OrpazFishel, line 96: Consider moving it to where you use it.

src/main/java/com/google/research/bleth/servlets/NewExperimentSimulationServlet.java
```diff
@@ -0,0 +1,138 @@
...
+        double transmissionThresholdRadius = Double.parseDouble(request.getParameter("transmissionThresholdRadius"));
+
+        // Mapping strategies string to types.
+        StrategiesMapper strategiesMapper = StrategiesMapper.getInstance();
+        MovementStrategyFactory.Type beaconMovementStrategy = strategiesMapper.getMovementStrategy(beaconMovementStrategyAsString);
+        MovementStrategyFactory.Type observerMovementStrategy = strategiesMapper.getMovementStrategy(observerMovementStrategyAsString);
+        AwakenessStrategyFactory.Type observerAwakenessStrategy = strategiesMapper.getAwakenessStrategy(observerAwakenessStrategyAsString);
+
+        String responseText = "Simulation has been created successfully.";
+        String simulationId;
```
OrpazFishel, line 76: Why is it necessary?

src/main/java/com/google/research/bleth/servlets/NewExperimentSimulationServlet.java
```diff
@@ -0,0 +1,138 @@
...
+import java.io.IOException;
+import java.util.logging.Logger;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * A servlet used for creating and running a new simulation associated to an experiment,
+ * and updating experimentsToSimulations database entity.
```
OrpazFishel, line 40: Please explain what is ExperimentsToSimulations.

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
omermadmon: I just checked and it seems like it does required: e.g Double 10 will be converted into the string '10.0' and then the worker servlet won't be able to parse it using `Integer.parseInt()`.

OrpazFishel: I assume Double.parseDouble(...).intValue() can fix this, but I'm not sure if it's better than the current situation...

src/main/java/com/google/research/bleth/servlets/NewExperimentSimulationServlet.java
```diff
@@ -0,0 +1,125 @@
...
+                    .setAwakenessCycle(awakenessCycle)
+                    .setAwakenessDuration(awakenessDuration)
+                    .setTransmissionThresholdRadius(transmissionThresholdRadius)
+                    .build();
+
+            simulation.run();
+            updateExperiment(simulation.getId(), experimentId);
+        } catch (RuntimeException | EntityNotFoundException e) {
+            responseText = e.getMessage();
+            response.getWriter().println(responseText);
```
OrpazFishel, line 97: You're writing it after the catch.

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
```diff
@@ -0,0 +1,234 @@
...
+        Map<String, Number> properties = configuration.stream()
+                .collect(Collectors.toMap(PropertyWrapper::property, PropertyWrapper::value));
+
+        boolean positiveDimensions = properties.get(Schema.SimulationMetadata.rowsNum).intValue() > 0 &&
```
OrpazFishel, line 211: Consider change it to be legalDimensions etc. instead of positiveDimensions; legal and positive aren't necessarily the same (cycleGreaterOrEqualDuration for example).

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
```diff
@@ -0,0 +1,234 @@
...
+        double upper = Double.parseDouble(request.getParameter("upper" + upperCaseFirstChar(parameter)));
+        double step = Double.parseDouble(request.getParameter("step" + upperCaseFirstChar(parameter)));
+        Set<PropertyWrapper> values = new HashSet<>();
+        for (double value = lower; value <= upper; value += step) {
+            values.add(PropertyWrapper.create(parameter, value));
+        }
+        return ImmutableSet.copyOf(values);
+    }
+
+    private String upperCaseFirstChar(String s) {
```
nolaviz, line 172: Nit: the proper verb for converting to uppercase is "capitalize".
(Method name is usually a verb)

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
```diff
@@ -0,0 +1,234 @@
...
+        }
+        return String.join("&", keyValuePairs);
+    }
+
+    private boolean validateArguments(List<PropertyWrapper> configuration) {
+        Map<String, Number> properties = configuration.stream()
+                .collect(Collectors.toMap(PropertyWrapper::property, PropertyWrapper::value));
+
+        boolean legalDimensions = properties.get(Schema.SimulationMetadata.rowsNum).intValue() > 0 &&
+                properties.get(Schema.SimulationMetadata.rowsNum).intValue() > 0;
```
nolaviz, line 212: Should this be `colsNum`?

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
```diff
@@ -0,0 +1,234 @@
...
+    private String toQueryString(Map<String, String> params) throws UnsupportedEncodingException {
+        List<String> keyValuePairs = new ArrayList<>();
+        for (String param : params.keySet()) {
+            keyValuePairs.add(URLEncoder.encode(param, StandardCharsets.UTF_8.toString()) + "=" +
+                    URLEncoder.encode(params.get(param), StandardCharsets.UTF_8.toString()));
+        }
+        return String.join("&", keyValuePairs);
+    }
+
+    private boolean validateArguments(List<PropertyWrapper> configuration) {
```
nolaviz, line 207: Don't you already have a similar validation for simulation creation? (How _do_ you validate user-submitted simulations?)

src/main/java/com/google/research/bleth/servlets/NewExperimentSimulationServlet.java
```diff
@@ -0,0 +1,124 @@
...
+ */
+@WebServlet("/new-experiment-simulation")
+public class NewExperimentSimulationServlet extends HttpServlet {
+    private static final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    private static final Logger log = Logger.getLogger(NewExperimentSimulationServlet.class.getName());
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+        log.info("Experiment Simulation Worker Servlet invoked.");
```
nolaviz, line 50: Why do you need this log line?
You can search for this servlet in the logs by just querying for `protoPayload.resource="/new-experiment-simulation"`.

src/main/java/com/google/research/bleth/servlets/NewExperimentSimulationServlet.java
```diff
@@ -0,0 +1,124 @@
...
+        String beaconMovementStrategyAsString = request.getParameter("beaconMovementStrategy");
+        String observerMovementStrategyAsString = request.getParameter("observerMovementStrategy");
+        String observerAwakenessStrategyAsString = request.getParameter("observerAwakenessStrategy");
+        int awakenessCycle = Integer.parseInt(request.getParameter("awakenessCycle"));
+        int awakenessDuration = Integer.parseInt(request.getParameter("awakenessDuration"));
+        double transmissionThresholdRadius = Double.parseDouble(request.getParameter("transmissionThresholdRadius"));
+
+        // Mapping strategies string to types.
+        StrategiesMapper strategiesMapper = StrategiesMapper.getInstance();
+        MovementStrategyFactory.Type beaconMovementStrategy = strategiesMapper.getMovementStrategy(beaconMovementStrategyAsString);
```
nolaviz, line 71: Can this fail? How will your system behave if it does?
(Same applies to the next 3 lines)

nolaviz (changes requested):

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
```diff
@@ -0,0 +1,234 @@
...
+
+        try {
+            experiment = datastore.get(experimentId);
+            experiment.setProperty(Schema.Experiment.simulationsLeft, legalConfigurationsCount);
+            datastore.put(experiment);
+            log.info("Experiment entity with id " + KeyFactory.keyToString(experimentId) +
+                    "was updated with simulationsLeft=" + legalConfigurationsCount);
+        } catch (EntityNotFoundException e) {
+            response.setContentType("text/plain;");
+            response.getWriter().println(e.getMessage());
```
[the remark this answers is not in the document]
omermadmon: I actually think we can add the entity's `simulationsLeft` property while it is being created (right after line 85) instead of having this try-catch block (the reason this try-catch block exists is that previously the arguments validation occurred inside the for loop, and this number of legal configs was known only after the for loop - this is not the case anymore). 

What do you think?
@nolaviz @OrpazFishel

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
```diff
@@ -0,0 +1,234 @@
...
+
+        int legalConfigurationsCount = configurations.size();
+        for (List<PropertyWrapper> configuration : configurations) {
+            AppEngineHttpRequest httpRequest = toHttpRequest(configuration, beaconsNum, experimentId, experimentTitle);
+            log.info("A new AppEngineHttpRequest was created: " + httpRequest.toString());
+            enqueueTask(httpRequest);
+            log.info("A new Task was created and enqueued.");
+        }
+
+        try {
```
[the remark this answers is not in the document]
omermadmon: If I move this try-catch block before the tasks creation (which I agree is much better) - can I just add the `simulationsLeft` property to the entity before calling `datastore.put()` and avoid this potential exception?

@nolaviz

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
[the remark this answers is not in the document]
nolaviz: SG

src/main/java/com/google/research/bleth/servlets/NewExperimentSimulationServlet.java
omermadmon: If at this point one of the strategies strings is not valid, a `null` value will be returned from `getMovementStrategy()` / `getAwakenessStrategy()`. Then the Builder's `build()` method will throw a `NullPointerException` with a message indicating the missing argument, e.g "No beacon movement strategy has been set.".

The strategies string are also being validated on client side (in the case where the simulation is not associated with an experiment) or on the enqueuing servlet (in our case - it only uses the default RANDOM strategy, which is valid).

The only change I can think of is modifying the `getMovementStrategy()` / `getAwakenessStrategy()` method to throw an exception rather than returning `null` in such case. If you believe this is necessary I will do so (I believe this belongs to a separate PR thought).

@nolaviz

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
omermadmon: The SimulationBuilder does have such method:

```java
void validateArguments() {
            checkArgument(rowNum > 0 && colNum > 0,
                    "Board dimensions must be positive.");
            checkArgument(beaconsNum > 0 && observersNum > 0,
                    "Number of beacons and number of observers must be positive.");
            checkArgument(awakenessCycle > 0 && awakenessDuration > 0,
                    "Both awakeness cycle and duration must be positive.");
            checkArgument(awakenessCycle >= awakenessDuration,
                    "Awakeness cycle must be greater or equal to duration.");
            checkArgument(transmissionThresholdRadius > 0,
                    "Transmission threshold radius must be positive.");
            checkArgument(maxNumberOfRounds > 0,
                    "Maximum number of rounds must be positive.");
            checkNotNull(beaconMovementStrategyType, "No beacon movement strategy has been set.");
            checkNotNull(observerMovementStrategyType, "No observer movement strategy has been set.");
            checkNotNull(awakenessStrategyType, "No awakeness strategy type has been set.");
        }
```

This method accesses the builder's fields so it cannot be used outside the class. 
I thought of implementing a utility validation function that will be used by both the builder and the servlet (and gets the parameters as a map, for example) - but then it became too messy: the builder and the servlet need to convert their fields/configuration into a map (or any other DS the utility function accepts as an argument) each time they want to use this utility method.

On the other hand I do understand why this is better to have both the builder and the servlet validation depend on the same generic method. I will modify it this way (unless you believe the potential code-complexity doesn't worth it).

@nolaviz

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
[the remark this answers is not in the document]
nolaviz: Yes.

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
nolaviz: Can you use a `SimulationBuilder` here, instead of 12 separate primitive variables?

omermadmon: How about something like:

```java
private boolean validateArguments(List<PropertyWrapper> configuration) {
        Map<String, Number> properties = configuration.stream()
                .collect(Collectors.toMap(PropertyWrapper::property, PropertyWrapper::value));

        try {
            new TracingSimulation.Builder()
                    .setMaxNumberOfRounds(properties.get(Schema.SimulationMetadata.roundsNum).intValue())
                    .setRowNum(properties.get(Schema.SimulationMetadata.rowsNum).intValue())
                    .setColNum(properties.get(Schema.SimulationMetadata.colsNum).intValue())
                    .setBeaconsNum(properties.get(Schema.SimulationMetadata.beaconsNum).intValue())
                    .setObserversNum(properties.get(Schema.SimulationMetadata.observersNum).intValue())
                    .setBeaconMovementStrategyType(defaultMovementStrategy)
                    .setObserverMovementStrategyType(defaultMovementStrategy)
                    .setAwakenessStrategyType(defaultAwakenessStrategy)
                    .setAwakenessCycle(properties.get(Schema.SimulationMetadata.awakenessCycle).intValue())
                    .setAwakenessDuration(properties.get(Schema.SimulationMetadata.awakenessDuration).intValue())
                    .setTransmissionThresholdRadius(properties.get(Schema.SimulationMetadata.transmissionThresholdRadius)
                            .doubleValue())
                    .build();
        } catch (Exception e) {
            return false;
        }
        return true;
    }
```

@nolaviz

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
```diff
@@ -0,0 +1,234 @@
...
+        double step = Double.parseDouble(request.getParameter("step" + upperCaseFirstChar(parameter)));
+        Set<PropertyWrapper> values = new HashSet<>();
+        for (double value = lower; value <= upper; value += step) {
+            values.add(PropertyWrapper.create(parameter, value));
+        }
+        return ImmutableSet.copyOf(values);
+    }
+
+    private String upperCaseFirstChar(String s) {
+        return s.substring(0, 1).toUpperCase() + s.substring(1);
```
nolaviz, line 173: Note/nit: we recommend to use `Ascii.toUpperCase(s)` rather than `s.toUpperCase()`, as the latter doesn't define a specific language (and may be wrong for some characters, e.g. Turkish "ı" (dotless "i").

src/main/java/com/google/research/bleth/servlets/EnqueueExperimentServlet.java
nolaviz: This works, yeah, but it's a bit ugly...
Consider leaving a TODO in place to refactor `TracingSimulation.Builder.validateArguments()` so it's usable in more contexts (but we can leave actual refactoring for later, for now).

nolaviz (approved):

PR #54, opened 2021-01-06
Write and Read Interval Stats to/from DB

omermadmon:


src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
```diff
@@ -18,9 +18,11 @@
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.ImmutableTable;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.Table;
+
```
nolaviz, line 25: Google recommended style does not separate java.* imports from other imports. https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing

(Applies to other files touched in this PR as well)

src/main/java/com/google/research/bleth/simulator/StatisticsState.java
```diff
@@ -135,9 +185,22 @@ public void writeBeaconsObservedStats() {
...
+                            Table<String, String, Double> beaconsObservedStats,
+                            LinkedListMultimap<Integer, ObservedInterval> beaconsObservedIntervals) {
         this.simulationId = simulationId;
         this.distanceStats = distanceStats;
         this.beaconsObservedStats = beaconsObservedStats;
+        this.beaconsObservedIntervals = beaconsObservedIntervals;
+    }
+
+    private static ObservedInterval extractObservedInterval(Entity entity) {
+        int start = ((Long) entity.getProperty(Schema.StatisticsState.intervalStart)).intValue();
```
nolaviz, line 197: Do you really need this trip through `Long` --> `intValue()`? Couldn't you just cast directly to `Integer`?

src/main/java/com/google/research/bleth/simulator/StatisticsState.java
```diff
@@ -41,10 +46,30 @@
...
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+        for (Integer beaconId : beaconsObservedIntervals.keySet()) {
+            for (ObservedInterval interval : beaconsObservedIntervals.get(beaconId)) {
+                Entity entity = new Entity(Schema.StatisticsState.entityKindBeaconsObservedIntervals);
+                entity.setProperty(Schema.StatisticsState.simulationId, simulationId);
+                entity.setProperty(Schema.StatisticsState.beaconId, beaconId);
+                entity.setProperty(Schema.StatisticsState.intervalStart, interval.start());
+                entity.setProperty(Schema.StatisticsState.intervalEnd, interval.end());
+                entity.setProperty(Schema.StatisticsState.intervalObserved, interval.observed() ? 1 : 0);
```
nolaviz, line 69: Why are you using 1/0, and not something more descriptive (e.g. a string "OBSERVED" / "NOT_OBSERVED")?

src/main/java/com/google/research/bleth/simulator/StatisticsState.java
omermadmon: According to [setProperty](https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/PropertyContainer.html#setProperty-java.lang.String-java.lang.Object-) documentation:

> As the value is stored in the datastore, it is converted to the datastore's native type. This may include widening, such as converting a Short to a Long. 

Also discussed [here](https://stackoverflow.com/questions/15246265/calling-setproperty-with-int-but-getproperty-returns-long-on-google-app-engine-p).

src/main/java/com/google/research/bleth/simulator/AbstractSimulation.java
omermadmon: As discussed, will be fixed in a separate PR.

nolaviz (approved):

